PR #3, opened 2016-02-09
[BC break] Pass events as an option instead of a required argument.

n1k0:
Atm if you want to have a local EventEmitter instance created for you and provide custom options, you need to write something like this:

```js
const api = new Api("http://foo", undefined, {requestMode: "cors"})
```

There's no point at providing an undefined second arg, so we should just do:

```js
const api = new Api("http://foo", {requestMode: "cors"})
```

And if we want to provide a custom event emitter:

```js
const events = new EventEmitter();
const api = new Api("http://foo", {events, requestMode: "cors"})
```

r=? @leplatrem

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -21,14 +21,15 @@ export default class KintoApi {
    * Constructor.
    *
    * Options:
-   * - {Object} headers      The key-value headers to pass to each request.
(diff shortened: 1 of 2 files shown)


leplatrem:
r+

PR #4, opened 2016-02-09
Initial integration test setup

n1k0:
This sets up the integration environment using a real Kinto server.

Next steps are to extract the `KintoServer` test utility class to its own package so we can conveniently reuse it in Kinto.js as a dev dependency.

r=? @leplatrem


n1k0:
r+

PR #5, opened 2016-02-10
Add get_records() method (pair=@n1k0)

leplatrem:
This allows retrieving the list of records for a given collection from a given bucket.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -325,13 +325,73 @@ export default class KintoApi {
   }
 
   /**
+   * Creates a new bucket.
+   *
+   * Options:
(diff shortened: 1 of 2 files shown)


n1k0:
Question: would it make sense to pass the bucket name as an option? Many users will use Kinto's default bucket.

```js
// default bucket
api.getRecords("default", "tasks");
// custom bucket
api.getRecords("todos", "tasks");
```

vs.


```js
// default bucket
api.getRecords("tasks");
// custom bucket
api.getRecords("tasks", {bucket: "todos"});
```
Thoughts?

leplatrem:
r+

PR #6, opened 2016-02-11
Refactored batch API and global architecture

n1k0:
This is now ready for preliminary review.

r=? @leplatrem

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,9 +3,11 @@
 import "isomorphic-fetch";
 import { EventEmitter } from "events";
 
-import { quote, unquote, partition } from "./utils.js";
+import { quote, unquote, partition, pMap } from "./utils.js";
 import HTTP from "./http.js";
-
+import endpoint from "./endpoint";
+import * as requests from "./requests";
+import { createBatch, aggregate } from "./batch";
 
 /**
  * Currently supported protocol version.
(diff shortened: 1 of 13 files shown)


src/index.js
```diff
@@ -204,124 +184,89 @@ export default class KintoApi {
...
+      headers: {
+        ...this.optionHeaders,
+        ...headers
       }
     });
-    return results;
+    fn(batch);
+    return this._batchRequests(batch.requests, options)
+      .then((responses) => {
+        if (options.aggregate) {
```
leplatrem: nit: `groupByStatus` ?

test/api_test.js
```diff
@@ -374,186 +313,167 @@ describe("Api", () => {
...
-        it("should map create & update requests", () => {
-          expect(requestBody.requests[0]).eql({
-            body: {
-              data: { id: 1, title: "foo" },
-            },
-            headers: { "If-Match": quote(42) },
-            method: "PUT",
-            path: `/${SPV}/buckets/blog/collections/articles/records/1`,
+        // XXX: batch() and requests should allow dealing with the If-Match
+        // header through some "safe" option. Keeping old tests skipped there.
```
leplatrem: I don't understand this comment.

n1k0: Forgot to push a commit, fixed by f991e05289398988150417500a4a95d62c02cc71

leplatrem:
I have one concern left :)

The signature of the `createRecord()` method.

Currently, we do that: 
```js
api.batch(batch => {
  batch.createBucket("custom");
  batch.createCollection("blog");
  batch.createRecord("blog", {title: "art1"});
  batch.createRecord("blog", {title: "art2"});
}, {bucket: "custom"})
```

I believe the `collection` should be an option instead (for consistency with bucket):

```js
api.batch(batch => {
  batch.createBucket("custom");
  batch.createCollection("blog");
  batch.createRecord({title: "art1"});
  batch.createRecord({title: "art2"});
}, {bucket: "custom", collection: "blog"})```

**Comment by n1k0** [comment]
> I believe the collection should be an option instead (for consistency with bucket):

I agree. Makes me think we could also define default options for these at the constructor level:

```js
const api = new Api("http://...", {bucket: "custom", collection: "blog"});
api.createRecord({title: "foo"});
```

What do you think?

leplatrem:
> What do you think?

I like it that way. It makes it very convienient to use.

Regarding state about default options for methods, I find it acceptable. But I may not be the most demanding person regarding that matter

n1k0:
Sold then, I'll update the patch accordingly.

n1k0:
I'm just realizing that while there's a `default` bucket, there's no `default` collection. So I think passing the collection name as a required argument still makes sense.

Also, passing a required value as an *option* is highly confusing.

Thoughts @leplatrem?

leplatrem:
>  passing a required value as an option is highly confusing.

True.

r+ then :)

PR #10, opened 2016-02-18
Added Api#updateRecord.

n1k0:
r=? @leplatrem


n1k0:
As most of the patch as been pair-reviewed already, I'm landing this.

Issue #14, opened 2016-02-20
Homogeneize and centralize how we deal with request options


n1k0:

n1k0:
Fixed by #15

PR #15, opened 2016-02-20
Better client request options handling.

n1k0:
Refs #14. r=? @leplatrem @almet 

Note: this also adds support for a default `safe` option which was not available at the instance level.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -26,6 +26,8 @@ export default class KintoApi {
    * - {EventEmitter} events      The events handler. If none provided an
    *                              `EventEmitter` instance will be created.
    * - {Object}       headers     The key-value headers to pass to each request.
+   * - {Boolean}      safe        Adds concurrency headers to every requests
+   *                              (default: `false`).
    * - {String}       requestMode The HTTP request mode (from ES6 fetch spec).
    *
    * @param  {String} remote  The remote URL.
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -26,6 +26,7 @@ export default class KintoApi {
    * - {EventEmitter} events      The events handler. If none provided an
    *                              `EventEmitter` instance will be created.
    * - {Object}       headers     The key-value headers to pass to each request.
+   * - {Boolean}      safe        Perform safe requests (default: `false`).
```
almet: that would be great to explain what a "safe" request is.

test/api_test.js
```diff
@@ -90,6 +90,11 @@ describe("Api", () => {
       const api = new Api(`http://test/${SPV}`, {bucket: "custom"});
       expect(api.defaultBucket).eql("custom");
     });
+
+    it("should accept a safe option", () => {
+      const api = new Api(`http://test/${SPV}`, {safe: true});
```
almet: not really related to what we're doing here, but since we're using `http://test/${SPV}` in different places, should we actually define a `remote_url` const somewhere and use it?

almet:
Looks good to me, with or without the nitpicks addressed :)

src/index.js
n1k0: I think we'll have to create proper human readable for this project, esdocs are not enough I'm afraid. See #16.

test/api_test.js
n1k0: Addressed in 1275415b322a4fa9c0d6a4d4931469afc93cc719

PR #29, opened 2016-02-23
Fixes #26: Excluded internal methods from esdocs.

### LINKED ISSUE #26: Exclude internal methods from generated esdocs.
We should only promote and document the intended public API.

n1k0:
Refs #26


n1k0:
r=me

PR #37, opened 2016-02-28
Remove duplicated methods

n1k0:
This basically migrates individual bucket, collection and record management methods from KintoClient to Bucket and Collection classes, to avoid having their logic duplicated in different places

This also removes the previous `chain` module and moves its `Bucket` and `Collection` classes to their own respective modules.

Last, it adds support for passing options to `KintoClient#bucket()` and ensures they're propagated down the flow.

Atomic commits should help reviewing this patch.


n1k0:
This is ready for review, but I suspect this is too big to motivate anybody.

We can probably live pair review this. r=? @leplatrem @almet @Natim

Natim:
I have read all the code, I am ok for a little explanation :)

n1k0:
> I have read all the code, I am ok for a little explanation :)

For the records, here's the resulting IRC discussion:

```
16:55:02 <NiKo`> natim: just read your comment on my PR, is it you disagree with the changes, or just don't understand them?
16:55:34 <natim> I don't disagree
16:55:48 <natim> I am not sure what it is about
16:56:23 <NiKo`> avoiding to allow doing the same things in several different ways
16:57:42 <NiKo`> before the patch eg. you could do client.createRecord("posts", {title: "foo"}, {bucket: "blog"}) and client.bucket("blog").collection("posts").createRecord({title: "foo"})
16:57:49 <NiKo`> now only the latter is supported
16:58:04 <NiKo`> and same for other actions
16:58:04 <natim> Oh really?
16:58:10 <natim> Is it something we want?
16:58:16 <NiKo`> this removes a bunch of things to maintain as well
16:58:18 <natim> I though we changed for the former
16:58:21 <NiKo`> yeas we do
16:58:26 <NiKo`> well, at least, I do ;)
16:59:01 <NiKo`> just for the sakje of clarity, this is *not* a stateful api, even if it's chained
16:59:21 <natim> Does it means that I can write:
16:59:27 <natim> collection = client.bucket("blog").collection("posts")
16:59:35 <NiKo`> the returned instances only cares about the parent resource to build endpoint urls, and handle default options like safe and friends
16:59:35 <natim> collection.createRecord({title: foo})
16:59:41 <NiKo`> yes
16:59:45 <natim> Ok perfect then
```

src/bucket.js
```diff
@@ -0,0 +1,174 @@
+import { toDataObj } from "./utils";
```
leplatrem: nit: `toDataBody`?

src/bucket.js
```diff
@@ -0,0 +1,174 @@
...
+
+
+  /**
+   * Retrieves bucket properties.
+   *
+   * @param  {Object} options         The options object.
+   * @param  {Object} options.headers The headers object option.
+   * @return {Promise<Object, Error>}
+   */
+  getProperties(options={}) {
```
leplatrem: nit: I think we call them *attributes* in the Cliquet/Kinto docs.

leplatrem:
I can't see any huge issue.

I didn't review the tests though.

n1k0:
Okay, I'm landing this.

PR #38, opened 2016-02-29
Added support for records list filtering.

n1k0:
r=? @leplatrem @Natim @almet @magopian (first in served first (please don't fight))

### DIFF
--- a/test/collection_test.js
+++ b/test/collection_test.js
@@ -424,6 +424,17 @@ describe("Collection", () => {
       return coll.listRecords()
         .should.become([{a: 1}]);
     });
+
+    describe("Filtering", () => {
+      it("should generate the expected filtering query string", () => {
+        coll.listRecords({sort: "x", filters: {min_y: 2, not_z: 3}});
+
+        sinon.assert.calledWithMatch(client.execute, {
+          path: "/buckets/blog/collections/posts/records?min_y=2&not_z=3&_sort=x",
+          headers: {Foo: "Bar", Baz: "Qux"},
+        });
+      });
+    });
   });
 
   /** @test {Collection#batch} */
--- a/test/collection_test.js
+++ b/test/collection_test.js
@@ -429,8 +429,19 @@ describe("Collection", () => {
       it("should generate the expected filtering query string", () => {
         coll.listRecords({sort: "x", filters: {min_y: 2, not_z: 3}});
 
+        const expectedQS = "min_y=2&not_z=3&_sort=x";
         sinon.assert.calledWithMatch(client.execute, {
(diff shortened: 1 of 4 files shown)


n1k0:
I'm getting mad at this:

n1k0:
Okay, I realize maybe the filter values needs url encoding or escaping; I'll check that out though if you have hints, that's cool too.

test/integration_test.js
```diff
@@ -18,7 +18,10 @@ const TEST_KINTO_SERVER = "http://0.0.0.0:8888/v1";
 describe("Integration tests", () => {
   let sandbox, server, api;
 
-  before(() => server = new KintoServer(TEST_KINTO_SERVER));
+  before(function() {
+    this.timeout(20000);
```
almet: ooh :(

test/integration_test.js
```diff
@@ -698,6 +714,22 @@ describe("Integration tests", () => {
...
+                return coll.batch(batch => {
+                  batch.createRecord({name: "paul", age: 28});
+                  batch.createRecord({name: "jess", age: 54});
+                  batch.createRecord({name: "john", age: 33});
+                });
+              });
+
+              it("should filter records", () => {
+                return coll.listRecords({sort: "age", filters: {min_age: 30}})
+                  .then(records => records.map(x => x.name))
```
almet: s/x/r ?

almet:
About the error you're getting: what do you think about having the client do a requests until it actually works (with an API built with promises, that could be a neat wait to avoid timeouts)

test/integration_test.js
leplatrem: I didn't open any ticket on this. But @n1k0 said that the last versions takes a lot longer to start

leplatrem:
Looks good from here :)

test/collection_test.js
```diff
@@ -424,6 +424,17 @@ describe("Collection", () => {
...
+
+    describe("Filtering", () => {
+      it("should generate the expected filtering query string", () => {
+        coll.listRecords({sort: "x", filters: {min_y: 2, not_z: 3}});
+
+        sinon.assert.calledWithMatch(client.execute, {
+          path: "/buckets/blog/collections/posts/records?min_y=2&not_z=3&_sort=x",
+          headers: {Foo: "Bar", Baz: "Qux"},
+        });
+      });
```
n1k0: Add a test without `sort` to ensure the qs is still properly generated.

n1k0:
> About the error you're getting: what do you think about having the client do a requests until it actually works (with an API built with promises, that could be a neat wait to avoid timeouts)

Yeah, that's actually what's done for flushing the server, I should probably port that strategy to server start as well. File #39 about that.

PR #41, opened 2016-03-01
Paginate records list

n1k0:
This patch adds support for paginating the list of records.

**Note:** This changes the API for `Collection#listRecords`, as it now requires to resolve with an Object exposing the pagination API instead of the list of records directly.

Excerpt from added docs:

#### Pagination

All records are retrieved by default. To specify a max number of records to retrieve, you can use the `limit` option:

```js
client.bucket("blog").collection("posts")
  .listRecords({limit: 20})
  .then(result => ...);
```

To fetch the following batch of results, just call `next()` from the result object obtained. If no next page is available, `next()` throws an error you can catch to exit the flow:

```js
client.bucket("blog").collection("posts")
  .listRecords({limit: 20})
  .then(({data, next}) => {
    try {
      return next();
    } catch(err) {
      return data;
    }
  });
  .then(records => ...);
```

Last, if you just want to retrieve and aggregate a given number of pages of results, instead of dealing with calling `next()` recursively you can simply specify the `pages` option:

```js
client.bucket("blog").collection("posts")
  .listRecords({limit: 20, pages: 3})
  .then(records => ...); // A maximum of 60 records will be retrieved here

### DIFF
--- a/README.md
+++ b/README.md
@@ -602,22 +602,26 @@ client.bucket("blog").collection("posts")
 Sample result:
 
 ```js
-[
-  {
-    "content": "True.",
-    "last_modified": 1456183930780,
-    "id": "a89dd4b2-d597-4192-bc2b-834116244d29",
-    "title": "I love cheese"
-  },
-  {
-    "content": "Yo",
-    "last_modified": 1456183914275,
-    "id": "63c1805a-565a-46cc-bfb3-007dfad54065",
-    "title": "Another post"
-  }
-]
-
+{
+  next: <Function>,
+  data: [
+    {
+      "content": "True.",
+      "last_modified": 1456183930780,
+      "id": "a89dd4b2-d597-4192-bc2b-834116244d29",
+      "title": "I love cheese"
+    },
+    {
+      "content": "Yo",
+      "last_modified": 1456183914275,
+      "id": "63c1805a-565a-46cc-bfb3-007dfad54065",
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by Natim** [comment]
Can we also have ``result.next()`` to raise if we do not have anymore Next-Page header?

**Comment by n1k0** [comment]
Oh, I debated this, actually. The API would be a little cumbersome methinks:

```js
client.bucket("blog").collection("posts")
  .listRecords({limit: 20})
  .then(result => result.next())
  .catch(err =>  
    console.log("Done paginating.")
  });
```

Natim:
```js
client.bucket("blog").collection("posts")
  .listRecords({limit: 20})
  .then(result => {
     try {
         return result.next()
     } catch (e) {
       return [];
     }
   })
```

n1k0:
Ah, I pretty much like it actually. Will give it a go, thanks :)

n1k0:
@Natim implemented your suggestion in 318878fb925ace9fb1fef65ebef2ba488e49e87f

n1k0:
@Natim added a `pages` option; thoughts?

Natim:
Yes maybe we should also support ``pages: all`` or ``allPages: true``

n1k0:
Or `pages: Infinity`. That's a good idea, I'll add support for this.

Natim:
:+1:

PR #52, opened 2016-03-04
Fixed support for If-None-Match: * headers for updates.

n1k0:
Currently the api when updating and deleting resources raise when the `safe` option is passed as `true` and no `last_modified` value is provided in the resource object or through the dedicate option.

It appears this only makes sense for deletions, but not for updates where one might want to ensure the resource won't be created automatically if it doesn't exist on the server.

This patch updates the code so it now only raises when asking for safe deletions without providing any `last_modified` information. 

r=? @Natim


test/requests_test.js
```diff
@@ -9,7 +9,7 @@ chai.should();
 chai.config.includeStack = true;
 
 
-describe("requests module", () => {
+describe.only("requests module", () => {
```
Natim: nit: only

test/integration_test.js
```diff
@@ -698,6 +698,15 @@ describe("Integration tests", function() {
...
               });
+
+              it("should not override existing data with no last_modified", () => {
+                return coll.createRecord({title: "foo"})
+                  .then(({data}) => coll.updateRecord({
+                    id: data.id,
+                    title: "foo",
+                  }, {safe: true}))
+                  .should.be.rejectedWith(Error, /412 Precondition Failed/);
+              });
```
n1k0: We should add a test to ensure a non-existent resource is created when safe is true.

leplatrem:
r+ !

n1k0:
v0.4.2 released https://github.com/Kinto/kinto-client/releases/tag/v0.4.2

PR #60, opened 2016-03-09
Fixes #55: Fetch server settings on first issued HTTP request.

### LINKED ISSUE #55: Ensure KintoClient#execute retrieves the server settings the first time it's called
Right now we only ensure server settings are retrieved and persisted in state when calling `batch`. We should rather ensure they're always retrieved first when issuing any first HTTP request to the server, and that should be done in `execute`.

n1k0:
I'm really challenging the benefits of this. Do we have an exhaustive list of exposed server settings documented somewhere, so we can accurately challenge future usefulness of this patch? /cc @leplatrem

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -200,7 +200,9 @@ export default class KintoClient {
     if (this.serverSettings) {
       return Promise.resolve(this.serverSettings);
     }
-    return this.execute({path: endpoint("root")})
+    return this.http.request(this.remote + endpoint("root"), {
+      headers: this.defaultReqOptions.headers
+    })
       .then(res => {
         this.serverSettings = res.json.settings;
         return this.serverSettings;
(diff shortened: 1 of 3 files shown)


leplatrem:
I am really not sure about this... Especially because you didn't remove it from `_batchRequest` so I don't see well the purpose of fetching (+reify) on each request.

Indeed, the root url contains a lot of useful information that kinto-client users might want to inspect at any time:
* `capabilities`
* user `id` and `bucket`
* protocol version `http_api_version`
* settings such as `readonly` and `max_batch_requests` (the list depend on external plugins)

So maybe instead of settings, we could reifiy the root url info.

n1k0:
> Especially because you didn't remove it from _batchRequest

Because we need to consume the batch_max_request setting, which can only be retrieved asynchronously the first time you need it and it's not available from instance state.

> So maybe instead of settings, we could reifiy the root url info. 

If you look at the issue tracker, I've already filed issues for that.

leplatrem:
Yes, so actually this PR could be reworded: fetch the settings on **the first** request.

I believe we could add a comment for the reader in `execute()` :)

r+ otherwise !

n1k0:
I'm actually thinking of revamping the settings retrieval API to always require it to be consumed asynchronously using a Promise:

```js
client.listSettings() // reads from cache or perform HTTP req
  .then(settings => ...) 

  client.getSetting("batch_max_requests") // reads from cache or perform HTTP req
    .then(setting => ...) 
```

Later we'll add the same kind of apis for retrieving capabilities, user, http_api_version and so on.

What do you think?

Natim:
> I'm actually thinking of revamping the settings retrieval API to always require it to be consumed asynchronously using a Promise. What do you think?

:+1: Really nice API.

PR #65, opened 2016-03-10
[Request for feedback] Investigations for capabilities checks.

almet:
Here is a WIP we worked on with @Natim this morning.

This relies on the ES6 Proxy feature, which isn't currently supported by node.
It seems possible to use it via Babel but it seems to make more sense to avoid
using ES6 proxies for now and do it "by hand".

However, what do you think of this approach? cc @n1k0 @leplatrem @magopian

### DIFF
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -1,6 +1,5 @@
 import { toDataBody } from "./utils";
 import Collection from "./collection";
-import * as requests from "./requests";
 import endpoint from "./endpoint";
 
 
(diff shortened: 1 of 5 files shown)


n1k0:
I'd personally rather go the decorators route, as described in https://github.com/Kinto/kinto-client/issues/62#issuecomment-194836624

n1k0:
Also, I'm a little confused with `deleteBuckets` and `deleteCollections` being listed as capabilities. Aren't these core features of the Kinto protocol? IMHO we should distinguish between third party provided capabilities (eg. plugin) and core features introduced by new minor bumps of the protocol. Am I one more time misunderstanding everything or does it make sense?

almet:
Yes that's correct, these aren't "capabilities", and we should find a better naming for them, but that's just so you can get the gist of the proposition (naming put aside)

n1k0:
Right. I'm gonna quickly draft what usage of decorators could look like in a next PR, so we'll have another point for comparison.

almet:
Yeah :)

leplatrem:
Superseed by #66

PR #66, opened 2016-03-10
Fixes #62: Check for http_api_version compatibility.

### LINKED ISSUE #62: Handle minor changes between kinto protocol versions
Currently, minor version changes in the Kinto protocol can introduce new features (e.g. a new delete buckets endpoint).

We can end up in a state where a new Kinto client (e.g. supporting Kinto protocol v1.4) tries to talk with a server which supports an older version of the protocol (e.g. v1.3), resulting in errors (e.g. HTTP 405 if the method doesn't exist on the endpoint, etc.).

To avoid having this behaviour, it is possible to have the client check if the operation is supported by the server (by comparing the server and client protocol versions).

n1k0:
This is WiP, refs #62.

The patch introduces a `ensureSupported` method at the client level we can leverage to conditionally execute a function depending of HTTP API version requirements.

This also adds `KintoClient#deleteBuckets()` which is provided by the Kinto protocol v1.4 only, and conditional integration tests for it. 

Feedback=? @almet @Natim @leplatrem @magopian 

PS: Discarded using ES7 decorators. I really wish decorated methods could access their related instance state, though we're only passed the class itself, which is pointless except if you want to attach static properties to it (which you don't). Also, fought too hard with Babel to be confident using them securely.

Edit: Finally managed to use ES7 decorators. Much better.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -391,4 +391,19 @@ export default class KintoClient {
     return this.execute(requests.deleteBucket(_bucket, reqOptions))
       .then(res => res.json);
   }
+
+  /**
+   * Deletes all buckets on the server.
+   *
+   * @ignore
+   * @param  {Object}        options         The options object.
+   * @param  {Boolean}       options.safe    The safe option.
+   * @param  {Object}        options.headers The headers object option.
+   * @return {Promise<Object, Error>}
+   */
+  deleteBuckets(options={}) {
+    const reqOptions = this._getRequestOptions(options);
+    return this.execute(requests.deleteBuckets(reqOptions))
+      .then(res => res.json);
+  }
 }
--- a/src/index.js
+++ b/src/index.js
@@ -3,7 +3,7 @@
 import "isomorphic-fetch";
 import { EventEmitter } from "events";
 
-import { partition, pMap, omit } from "./utils";
+import { partition, pMap, omit, checkVersion } from "./utils";
 import HTTP from "./http";
 import endpoint from "./endpoint";
 import * as requests from "./requests";
@@ -149,6 +149,25 @@ export default class KintoClient {
     this.events.on("backoff", backoffMs => {
       this._backoffReleaseTime = backoffMs;
     });
+  }
+
+  /**
+   * Conditionnaly executes provided function if current server http_api_version
+   * supports it and returns a Promise resolving with its result when the
+   * condition is satisfied, rejects when it's not.
+   *
+   * @ignore
+   * @param  {String}   min  The minimum protocol version required (inclusive).
+   * @param  {String}   max  The maximum protocol version required (exclusive).
+   * @param  {Function} fn   The function to resolve with.
+   * @return {Promise<Object, Error>}
+   */
+  ensureSupported(min, max, fn) {
+    return this.fetchHTTPApiVersion()
+      .then(currentVersion => {
+        checkVersion(currentVersion, min, max);
+        return fn();
+      });
   }
 
   /**
--- a/src/index.js
+++ b/src/index.js
@@ -3,7 +3,7 @@
 import "isomorphic-fetch";
 import { EventEmitter } from "events";
 
-import { partition, pMap, omit, checkVersion, support } from "./utils";
+import { partition, pMap, omit, support } from "./utils";
 import HTTP from "./http";
 import endpoint from "./endpoint";
 import * as requests from "./requests";
@@ -149,22 +149,6 @@ export default class KintoClient {
     this.events.on("backoff", backoffMs => {
       this._backoffReleaseTime = backoffMs;
     });
-  }
-
-  /**
-   * Checks if current server http_api_version matches the provided min and max
-   * version requirements; returns a Promise resolving when the version is
-   * within range, rejecting when it's not.
-   *
-   * @ignore
-   * @param  {String}   min  The minimum protocol version required (inclusive).
-   * @param  {String}   max  The maximum protocol version required (exclusive).
-   * @param  {Function} fn   The function to resolve with.
-   * @return {Promise<Object, Error>}
-   */
-  ensureSupported(min, max) {
-    return this.fetchHTTPApiVersion()
-      .then(version => checkVersion(version, min, max));
   }
 
   /**
--- a/src/index.js
+++ b/src/index.js
@@ -152,9 +152,9 @@ export default class KintoClient {
   }
 
   /**
-   * Conditionnaly executes provided function if current server http_api_version
-   * supports it and returns a Promise resolving with its result when the
-   * condition is satisfied, rejects when it's not.
+   * Checks if current server http_api_version matches the provided min and max
+   * version requirements; returns a Promise resolving when the version is
+   * within range, rejecting when it's not.
    *
    * @ignore
    * @param  {String}   min  The minimum protocol version required (inclusive).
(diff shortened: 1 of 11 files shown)


n1k0:
i'm gonna implement `deleteBuckets()` and `deleteCollections()` to leverage this as a PoC.

n1k0:
So writing conditional tests is done that way: https://github.com/Kinto/kinto-client/blob/support-checks/test/integration_test.js#L159-L181

It's not ideal, though quite okay so we could live with this for quite some time without too much hassle I suppose.

Still wanting feedback before proceeding further here.

src/index.js
```diff
@@ -372,4 +388,22 @@ export default class KintoClient {
...
+   * Deletes all buckets on the server.
+   *
+   * @ignore
+   * @param  {Object}  options         The options object.
+   * @param  {Boolean} options.safe    The safe option.
+   * @param  {Object}  options.headers The headers object option.
+   * @return {Promise<Object, Error>}
+   */
+  deleteBuckets(options={}) {
+    return this.ensureSupported("1.4", "2.0")
```
leplatrem: Well, but that would mean implicit `ensureSupported(1.3, 2.0)` everywhere ?

n1k0: No, that means we'll have to update everything to explicitly use this method for every method. And yes, I really wish we could use decorators for that ;(

leplatrem: > Is this seriously a big issue? We can always revisit the max version at the time this next major of the protocol is out...

After thinking about it, yes, this is it! We can bump the next supported major whenever :) 

Sorry, I replied to this comment without going very deeply :( I had no intention to confuse you!

src/index.js
```diff
@@ -372,4 +372,20 @@ export default class KintoClient {
...
+   * Deletes all buckets on the server.
+   *
+   * @ignore
+   * @param  {Object}  options         The options object.
+   * @param  {Boolean} options.safe    The safe option.
+   * @param  {Object}  options.headers The headers object option.
+   * @return {Promise<Object, Error>}
+   */
+  @support("1.4", "2.0")
+  deleteBuckets(options={}) {
```
leplatrem: :+1: x 1000

n1k0:
Well, woops.



Tomorrow.

n1k0:
Huh, restarted the failing job and now it passes. Intermittent? Weird.

src/utils.js
```diff
@@ -86,3 +86,58 @@ export function toDataBody(value) {
...
+
+/**
+ * Checks if a version is within the provided range.
+ *
+ * @param  {String} version The version to check.
+ * @param  {String} min     The minimum version (inclusive).
+ * @param  {String} max     The minimum version (exclusive).
+ * @throws {Error} If the version is outside of the provided range.
+ */
+export function checkVersion(version, min, max) {
```
almet: nit: rename `min` and `max` to `minSupported` and `maxSupported`

src/utils.js
```diff
@@ -86,3 +86,58 @@ export function toDataBody(value) {
...
+  const extract = (str) => str.split(".").map(x => parseInt(x, 10));
+  const [verMajor, verMinor] = extract(version);
+  const [minMajor, minMinor] = extract(min);
+  const [maxMajor, maxMinor] = extract(max);
+  const msg = `Version ${version} doesn't match ${min} <= x < ${max}`;
```
almet: nit: put the error message just before the throw.

src/utils.js
n1k0: really? -_-

src/utils.js
```diff
@@ -86,3 +86,58 @@ export function toDataBody(value) {
...
+ * @return {Function}
+ */
+export function support(min, max) {
+  return function(target, key, descriptor) {
+    const fn = descriptor.value;
+    return {
+      configurable: true,
+      get() {
+        const wrappedMethod = (...args) => {
+          return this.fetchHTTPApiVersion()
```
almet: This makes me realize the `support` decorator is used on `index.js` (the client), meaning that batch operations won't benefit from the check.

Should we do this on the request object so that batch operations also issue a warning ? (see #65 )

test/integration_test.js
```diff
@@ -155,6 +156,30 @@ describe("Integration tests", function() {
       });
     });
 
+    describe("#deleteBuckets()", () => {
+      before(function() {
+        try {
+          checkVersion(server.http_api_version, "1.4", "2.0");
+        } catch(err) {
+          this.skip();
```
almet: skip? Why?

almet: (We should probably issue an error in this case)

src/index.js
```diff
@@ -372,4 +372,20 @@ export default class KintoClient {
...
+  /**
+   * Deletes all buckets on the server.
+   *
+   * @ignore
+   * @param  {Object}  options         The options object.
+   * @param  {Boolean} options.safe    The safe option.
+   * @param  {Object}  options.headers The headers object option.
+   * @return {Promise<Object, Error>}
+   */
+  @support("1.4", "2.0")
```
almet: nit: rename to `supports` (with the extra "s") ?

src/utils.js
almet: That's your call, but I had to see the examples to understand what these min and max versions were about. Maybe just in the esdoc ?

almet:
This looks great! I love the simplicity of the implementation here: it's really easy to add this feature to already existing and new endpoints, and make it very explicit to the potential reader which versions are supported by each operation.

n1k0:
I managed to reproduce the intermittent failure I had previously on Travis, it seems the server can take a lot of time to delete all buckets; more worrying, it seems it sends a success response even if the operation is not entirely done just yet.

Is this expected or should I file an issue about this? @almet @leplatrem @Natim @magopian

test/integration_test.js
n1k0: No, this is a test ran contextually depending on the http api version used by the server, as we decided we wanted to go that route.

Basically, before running the test, it checks for the protocol version used by the server, and depending on it will run or skip running the test.

src/index.js
n1k0: I don't feel a huge need for it.

src/utils.js
n1k0: That's a good point, and fixing this isn't trivial at all, as requests are agnostic from the http_api_version retrieval procedure... I need to think a little more about that.

almet:
So, it depends which status you're getting back. The HTTP spec says that:

- A 200 means that "The request has succeeded"
- A 202 means that the request was accepted but not yet fulfilled;
- A 204 means that "The server has fulfilled the request" but there is nothing to answer (that's what's usually used on DELETE operations)

In cliquet, [it seems that we return a 200](https://github.com/mozilla-services/cliquet/blob/master/cliquet/resource/__init__.py#L499), which doesn't seem to be consistent with what you're describing here.

Please, open an issue about that, yes!

n1k0:
Filed https://github.com/mozilla-services/cliquet/issues/682 about that.

n1k0:
For checking versions at the request level for `batch` support, I'll file another issue. I badly want to land this and iterate.

almet:
> I badly want to land this and iterate.
Please do!

src/index.js
almet: Tt seems more grammatically correct :-)

almet: (E.g. that reads better: "this feature supports version 1.4 to 2.0")

leplatrem:
> it seems it sends a success response even if the operation is not entirely done just yet.

I can't believe that :)

src/index.js
n1k0: Hmm would be read as well as *support for this feature is provided from version 1.4 to 2.0*...

n1k0:
I'm landing this.

almet:
Great!

PR #72, opened 2016-03-12
[Experiment] Revamped batch API.

n1k0:
This experimental patch revamps the batch API, removing the previous `createBatch` convoluted implementation and associated hacks, switching to use a specifically configured `KintoClient` instance instead as a batch object, so it exposes the very same API within a batch.

This will allow defining version support in a single place, and could probably invite moving all request generation in all the methods directly, as we don't really need to have them decoupled for batching anymore.

Feedback welcome.


src/bucket.js
```diff
@@ -71,6 +81,7 @@ export default class Bucket {
    * @param  {Object} options.headers The headers object option.
    * @return {Promise<Object, Error>}
    */
+  @nobatch("Batch doesn't support read operations.")
```
Natim: Why not?

n1k0: Because the result of a read operation within a batch would be sent along all the responses received from the main batch one, and you usually want to consume directly what you're reading.

Natim: What about something like that:

```es6
api.bucket("gartenbilder").batch(batch => {
    batch.collections('albums').get_records();
    batch.collections('pictures').get_records();
});
```

n1k0: Please reread my comment above?

Natim: Isn't it the same for all of them:

https://github.com/Kinto/kinto-client/pull/72/files#diff-d2dbe7779d59d4dd594e92248caba0a5R245

Why would you prevent people of doing a read operation in a batch if they plan on doing so?

Natim: I don't think the client should prevent something that the server accept as valid.

n1k0: How would you consume the result of the read operation in the middle of a synchronous batch description as you suggested above? Please provide an example use case (eg. ask yourself "when do I get the actual results from my read operations?")

n1k0: Hint 1:

```js
api.bucket("blog").batch(batch => {
    const records = batch.collections('articles').get_records(); // doesn't work, reading is async, not sync
    // ...
})
  .then((batchResults) => {
    // oh my, it's here I'm getting the read operation results, not above where I really needed them!
  });
```

Hint 2: JavaScript isn't Python.

Natim: > you won't be able to consume these results before having actually sent the requests asking for them

Yes I am aware of that.

If you just want to retrieve data and you do not need to process this data to add request to the batch then you can just do what you said:

```
api.bucket("blog").batch(batch => {
    batch.collections('articles').get_records(id="e9fe125c-e865-11e5-bbab-04dad2bf2cac");
    batch.collections('comments-e9fe125c-e865-11e5-bbab-04dad2bf2cac').get_records();
})
.then((batchResults) => {
    // oh my, it's here I'm getting the read operation results, not above where I really needed them!
    const articleResponse = batchResults[0];
    const commentsResponse = batchResults[1];
  });
```

n1k0: I see... I'm super worried people will try to consume the results within the batch function, though that's probably matter of documentation. Hope you'll help writing these ;)

leplatrem:
I find it sound and clean!

Plus this is the strategy implemented in `kinto.py` :)

GG!

PR #75, opened 2016-04-07
Add a directive to build an ES6 version for FFx

leplatrem:
* [x] Test together with https://github.com/Kinto/kinto.js/pull/384
* [ ] Wait for kinto.js 2.0 to be released
* [ ] Land with https://github.com/leplatrem/gecko-dev/pull/8


fx-src/index.js
```diff
@@ -0,0 +1,35 @@
...
+const Cu = Components.utils;
+
+Cu.importGlobalProperties(['fetch']);
+const { EventEmitter } = Cu.import("resource://devtools/shared/event-emitter.js", {});
+
+export default class KintoHttpClient extends KintoClient {
+  constructor(remote, options={}) {
+    const emitter = {};
+    EventEmitter.decorate(emitter);
+    super({...options, events: emitter});
```
n1k0: We shouldn't override `events` by default here, as it's possibly been instantiated already and passed to the constructor, and we probably want to preserve it.

leplatrem:
I tried to use the `--bare` option to remove node shims from the build.

Unfortunately I run into this error when using it in Gecko:
```
 0:01.46 LOG: Thread-1 ERROR Error: Cannot find module 'events' at resource://services-common/kinto-http-client.js:26
s@resource://services-common/kinto-http-client.js:26:523
s/<@resource://services-common/kinto-http-client.js:26:680
[9]<@resource://services-common/kinto-http-client.js:952:15
```

I tried to run browserify with the `--ignore-missing` option, without success.

I don't explain it totally since we already have `--ignore events`. It should behave the same way as `--ignore isomorphic-fetch` which does not fail when required. Maybe because `events` is node native module?

n1k0:
Weird. What does using `--no-builtins` give?

n1k0:
> I'm curious as to see the bundled code trying to require this module...

Bah, we require it in `src/index.js`, hence the failure you get... We need to find another strategy.

n1k0:
I've pushed a fix in 4400bb5cf3453f59c3e3ba21a5708ac534a0ea63

Now sleep :sleeping:

leplatrem:
`events` problem is solved.
Now, we have to tackle `querystring`:
```
0:02.29 LOG: Thread-1 ERROR Error: Cannot find module 'querystring' at resource://services-common/kinto-http-client.js:26
s@resource://services-common/kinto-http-client.js:26:523
s/<@resource://services-common/kinto-http-client.js:26:680
[10]<@resource://services-common/kinto-http-client.js:1689:20
```

PR #85, opened 2016-05-26
Propagate and document timeout option (fixes #84)

### LINKED ISSUE #84: Document timeout option
The HTTP class supports a `timeout` option. We should document how to leverage it :) (and maybe its limitation) 

See https://github.com/Kinto/kinto.js/issues/280

leplatrem:
@n1k0 r?

### DIFF
--- a/src/base.js
+++ b/src/base.js
@@ -41,6 +41,7 @@ export default class KintoClientBase {
    * `"default"`)
    * @param  {String}  options.requestMode The HTTP request mode (from ES6 fetch
    * spec).
+   * @param  {Number}  options.timeout     The requests timeout in ms (default: `5000`).
    */
   constructor(remote, options={}) {
     if (typeof(remote) !== "string" || !remote.length) {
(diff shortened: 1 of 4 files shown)


src/base.js
```diff
@@ -86,12 +87,13 @@ export default class KintoClientBase {
...
     this.events = options.events;
 
+    const {requestMode, timeout} = options;
     /**
      * The HTTP instance.
      * @ignore
      * @type {HTTP}
      */
-    this.http = new HTTP(this.events, {requestMode: options.requestMode});
+    this.http = new HTTP(this.events, {requestMode, timeout});
```
Natim: Nice cleanup :+1:

n1k0:
r+

test/api_test.js
```diff
@@ -79,6 +79,16 @@ describe("KintoClient", () => {
         .eql(requestMode);
     });
 
+    it("should keep the default timeout in the child HTTP instance", () => {
+      expect(new KintoClient(sampleRemote).http.timeout).eql(5000);
```
Natim: A default timeout of 5s seems a little small to me, usually it is more something around 30s isn't it?

leplatrem: I left the existing value. 5 seconds is already quite a lot I think, no?

Natim: This PR is about the documentation so we can discuss that in another one if we think 5 is not enough.

PR #87, opened 2016-05-26
Document patch option (fixes #79)

### LINKED ISSUE #79: Document patch option for update methods
It fires a `PATCH` instead of `PUT`

leplatrem:
@Natim r?

### DIFF
--- a/README.md
+++ b/README.md
@@ -543,6 +543,7 @@ Sample result:
 
 #### Options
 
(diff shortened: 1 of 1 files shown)


n1k0:
r+

Natim:
Also if we were to keep ``setSchema`` it should be patch by default same for setData.

Issue #34, opened 2016-02-25
Handle Retry-After on HTTP 503


n1k0:
See what's done in kinto.py https://github.com/Kinto/kinto.py/blob/master/kinto_client/session.py#L69-L81

almet:
That would be a valuable addition for the heroku deploys as well: The requests are currently timing out because the worker is starting.

leplatrem:
Several solutions after a quick conversation with @n1k0 :

## minimal

* emit a `retry-after` event with the header value in payload.
Pros: would work like backoff
Cons: let the developers handle it, and possibly misbehave with what the server expects from clients.

## intermediary

Mimic python code.
* Add option N retry
* Wait N seconds and retry (`setTimeout`), where N initially comes from headers and then fibonacci to increase delay
* Raise error when retry exhausted
Pros: simple and assist the developper
Cons: keeps an event loop open, and several calls could retry in parallel

## nice

Same as intermediary but stack pending requests to avoid several calls to generate requests to the server while the *retry-after* is not elapsed.

PR #89, opened 2016-05-26
Minimal solution for retry-after (ref #34)

leplatrem:
@n1k0 r?

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -177,7 +177,7 @@ export default class HTTP {
     if (!retryAfter) {
       return;
     }
-    retryAfter = parseInt(retryAfter, 10);
+    retryAfter = (new Date().getTime()) + (parseInt(retryAfter, 10) * 1000);
     this.events.emit("retry-after", retryAfter);
   }
 }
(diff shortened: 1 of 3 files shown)


src/http.js
```diff
@@ -170,4 +171,13 @@ export default class HTTP {
...
     this.events.emit("backoff", backoffMs);
   }
+
+  _checkForRetryAfterHeader(status, headers) {
+    let retryAfter = headers.get("Retry-After");
+    if (!retryAfter) {
+      return;
+    }
+    retryAfter = parseInt(retryAfter, 10);
+    this.events.emit("retry-after", retryAfter);
```
n1k0: Nit: I wonder if we shouldn't explicitly mention the unit in the event name, eg. `retry-after-seconds`; yeah it's ugly but in my experience it prevents a lot of mistakes.

n1k0:
I like the minimal approach and the fact we delegate the handling of this to the end user :)

r+ with nit optionally addressed.

src/http.js
leplatrem: Yeah, I know what you mean... but damn it's ugly :(

We didn't have to do it for backoff because it provides a date instead of a delay. We should do the same for retry-after IMO.

Natim: We can convert this delay into date by using the current timestamp and adding the delay.

PR #90, opened 2016-05-26
Safety check on since value (fixes #51)

### LINKED ISSUE #51: listRecords "since" option should be validated as either string or undefined
Always better being notified early when you're about to send improper data to the server. Since the ETag is an opaque string value, we should only ensure it's either `undefined` (== option not provided) or a non-empty quoted string.

leplatrem:
@n1k0 r?

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -284,6 +284,10 @@ export default class Collection {
       sort: "-last_modified",
       ...options
     };
+    // Safety/Consistency check on ETag value.
+    if (since && typeof(since) != "string") {
+      throw new Error(`Invalid value for since (${since}), should be ETag value.`);
+    }
     const collHeaders = this.options.headers;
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -284,6 +284,10 @@ export default class Collection {
       sort: "-last_modified",
       ...options
     };
+    // Safety/Consistency check on ETag value.
+    if (since && typeof(since) != "string") {
```
n1k0: Out of curiosity, where could this happen?

leplatrem:
This can happen if a developper uses a record to filter with since `listRecords({since: myrecord.last_modified})`. Although it works technically speaking, it is recommended to use the `last_modified`value obtained from another call to `listRecords()` instead, as stated in the server docs.

leplatrem:
Oups, sorry, I was sure you had r+ :( I can revert if necessary.

PR #92, opened 2016-05-26
Rename methods for consistency (fixes #83)

### LINKED ISSUE #83: Consistent naming for collection and bucket methods operating on data
We should rename `getAttributes`, `setMetadata` and friends to just `getData` and `setData` and deprecate these old methods.

leplatrem:
@n1k0 please review thoroughly :)

* [x] `bucket.getData()` now returns `data`
* [x] rename attributes/metadata to data
* [x] remove get/setSchema
* [x] add missing documentation for get/setdata
* [x] Describe breaking changes

```
**Breaking changes**

- `client.bucket(...).getData()` now returns the `data` object instead of both data and permissions
- `Collection#setSchema()` and `Collection#getSchema()` were dropped. 
- `Collection#getAttributes()` was splited into `Collection#getData()` and `Collection#getPermissions()`
- `Collection#getMetadata()` is now `Collection#getData()`
- `Collection#setMetadata()` is now `Collection#setData()`
```

### DIFF
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -126,7 +126,7 @@ export default class Bucket {
    * @param  {Boolean} options.safe        The safe option.
    * @param  {Object}  options.headers     The headers object option.
    * @param  {Object}  options.permissions The permissions object.
-   * @param  {Object}  options.data        The metadadata object.
+   * @param  {Object}  options.data        The data object.
    * @param  {Object}  options.schema      The JSONSchema object.
    * @return {Promise<Object, Error>}
    */
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -127,7 +127,6 @@ export default class Bucket {
    * @param  {Object}  options.headers     The headers object option.
    * @param  {Object}  options.permissions The permissions object.
    * @param  {Object}  options.data        The data object.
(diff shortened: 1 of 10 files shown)


src/bucket.js
```diff
@@ -69,23 +69,24 @@ export default class Bucket {
...
    * @return {Promise<Object, Error>}
    */
-  getAttributes(options={}) {
+  getData(options={}) {
     return this.client.execute({
       path: endpoint("bucket", this.name),
       headers: {...this.options.headers, ...options.headers}
-    });
+    })
+    .then((res) => res.data);
```
n1k0: This is now the only method having this return value signature, I suspect this could be confusing to users; thoughts?

n1k0: Ok scratch this comment I've understood what you did.

src/utils.js
```diff
@@ -71,7 +71,7 @@ export function toDataBody(value) {
   if (typeof value === "string") {
     return {id: value};
   }
-  throw new Error("Invalid collection argument.");
+  throw new Error("Invalid argument.");
```
n1k0: nit: invalid entity argument

n1k0: nit2: s/value/entity (or resource?)

PR #93, opened 2016-05-26
setPermissions/setData bug (fixes #91)

### LINKED ISSUE #91: bucket.setPermissions() deletes existing data!
The request has both attribute `data` and `permissions`, but shouldn't.

And `data` contains `id` and `last_modified` only. The server will overwrite the existing data and remove any custom attribute defined on the bucket :) 

This might also be true for `setData()` which shouldn't overwrite existing permissions.

leplatrem:
* [x] Reproduce bug with integration tests (ref #91)
* [x] Wait for #83 to be merged
* [x] Fix code in requests.js
* [x] Wait for https://github.com/Kinto/kinto/pull/662 to be merged

### DIFF
--- a/src/requests.js
+++ b/src/requests.js
@@ -1,12 +1,13 @@
 import endpoint from "./endpoint";
+import { omit, removeUndefined } from "./utils";
 
 const requestDefaults = {
   safe: false,
   // check if we should set default content type here
   headers: {},
   bucket: "default",
-  permissions: {},
-  data: {},
+  permissions: undefined,
+  data: undefined,
   patch: false,
 };
 
@@ -33,7 +34,7 @@ export function createBucket(bucketName, options={}) {
     ...requestDefaults,
     ...options,
   };
-  return {
+  return removeUndefined({
     method: "PUT",
     path: endpoint("bucket", bucketName),
     headers: {...headers, ...safeHeader(safe)},
@@ -41,7 +42,7 @@ export function createBucket(bucketName, options={}) {
       data,
       permissions
     }
-  };
+  });
 }
 
 /**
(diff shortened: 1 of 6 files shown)


leplatrem:
This is ready to review @n1k0 r?

src/requests.js
```diff
@@ -20,28 +21,78 @@ function safeHeader(safe, last_modified) {
...
- */
-export function createBucket(bucketName, options={}) {
-  if (!bucketName) {
-    throw new Error("A bucket name is required.");
-  }
-  // Note that we simply ignore any "bucket" option passed here, as the one
-  // we're interested in is the one provided as a required argument.
-  const { data={}, headers, permissions, safe } = {
+
+function _create(path, {data, permissions}, options={}) {
```
n1k0: Nit: as we're already in a module and these functions are not exported, it's probably not necessary to prefix them with an underscore.

leplatrem: yeah but `delete` is a reserved keyword :/

n1k0: Oh, good point :o let's keep them that way then.

src/requests.js
```diff
@@ -20,28 +21,78 @@ function safeHeader(safe, last_modified) {
...
+    safe,
+    patch,
+  } = {...requestDefaults, ...options};
+  const { last_modified } = { ...data, ...options };
+
+  if (Object.keys(omit(data, "id", "last_modified")).length === 0) {
+    data = undefined;
+  }
+
+  return removeUndefined({
```
n1k0: I'm surprised this is needed as `undefined` properties are removed by serialization while it's already performed when sending the request body through `execute`: https://github.com/Kinto/kinto-client/blob/master/src/base.js#L366-L369

Am I missing something here? What breaks when this call is removed?

n1k0:
Looks good, just a remaining question I'd like to have an answer :)

leplatrem:
I'll merge this, because I took the comment into account in #99

PR #94, opened 2016-05-26
Add group support (fixes #33)

### LINKED ISSUE #33: Add support for permission groups.
Kinto docs: http://kinto.readthedocs.org/en/latest/api/1.x/groups.html

leplatrem:
* [x] bucket methods
* [x] add `getGroup()` method
* [x] add `listGroups()` method
* ~~Refactor to mimic collection: `group("admins").getData()`~~
* ~~Get rid of `updateGroup()`~~
* ~~Add highlevel fonctions like `addMember()` or `removeMember()` ?~~
* [x] Integration tests
* [x] documentation


leplatrem:
Is the current API ok or do we want this?

```js
client.bucket("blog").group("foo").getData()
```

```js
client.bucket("blog").group("foo").getPermissions()
```


```js
client.bucket("blog").group("foo").addMember("blah")
```


```js
client.bucket("blog").group("foo").removeMember("poo")
```

Natim:
I think it would make sense to have the ``.bucket("").group("")`` thing but for my usecase your solution is a good start I think.

leplatrem:
I am still doubtful about this actually...

Should groups be considered as *context objects* like buckets and collections, or like sub-object like records?

Reminder: there is no `.bucket("").record("...").getData()`.

Maybe group could just sub-objects of buckets, like records are sub-objects of collection. @n1k0 what do you think ?

n1k0:
Yeah I'm not the biggest fan of introducing a `.group` chainable API. We should aim at having the flattest and simplest API here.

Natim:
Oh ok I didn't get why we had .collection and .bucket now it makes sense. Yes getGroup is great :+

leplatrem:
This time, this is ready for review/merge. @n1k0 r?

I plan to refactor the whole requests module to generalize the create/update/delete to any object, in order to get rid of a lot of duplication in there. But I thought it would make sense to do it as a fix of #91

Natim:
Great \o/ r+

README.md
```diff
@@ -498,6 +503,184 @@ Sample result:
 - `safe`: Ensures the resource hasn't been modified in the meanwhile if `last_modified` is provided (default: `false`);
 - `last_modified`: The last timestamp we know the resource has been updated on the server.
 
+### Creating a group
+
```
n1k0: It may be worth shortly describing what a *group* is here, maybe linking to the kinto docs.

src/bucket.js
```diff
@@ -175,6 +175,106 @@ export default class Bucket {
...
+   */
+  createGroup(id, members=[], options={}) {
+    const reqOptions = this._bucketOptions(options);
+    const data = {
+      ...options.data,
+      id,
+      members
+    };
+    const path = endpoint("group", this.name, id);
+    const permissions = options.permissions;
```
n1k0: Nit: `const {permissions} = options;` (yes I'm such a hipster)

README.md
```diff
@@ -498,6 +503,184 @@ Sample result:
...
+#### Named group
```
n1k0: Shouldn't we mention implicitly named groups too?

src/bucket.js
```diff
@@ -175,6 +175,106 @@ export default class Bucket {
...
+    const request = requests.createRequest(path, {data, permissions}, reqOptions);
+    return this.client.execute(request);
+  }
+
+  /**
+   * Creates a new group in current bucket.
+   *
+   * @param  {String|undefined}  id        The group id.
```
n1k0: As far as I can tell this esdoc doesn't match the implementation at all :)

test/bucket_test.js
```diff
@@ -293,6 +293,220 @@ describe("Bucket", () => {
...
+        .getGroup("foo", {headers: {Baz: "Qux"}});
+
+      sinon.assert.calledWithMatch(client.execute, {
+        path: "/buckets/blog/groups/foo",
+        headers: {Foo: "Bar", Baz: "Qux"}
+      });
+    });
+
+    it("should return the group", () => {
+      getBlogBucket().getGroup("foo")
```
n1k0: Wait, I'm surprised this works as mocha expects a promise to be returned in order to perform the eventual assertion. Please double-check this test actually breaks if you change `fakeGroup`.

n1k0:
Looks super good. r+wc

src/bucket.js
leplatrem: You meant for `updateGroup()` right?

n1k0: Yup

n1k0:
final r+

PR #96, opened 2016-05-29
Fix default bucket update

n1k0:
Dunno if this should be considered a bug of kinto or kinto-client. When trying to post data to the default bucket, we get a 400 error if we pass the id as part of the data:

```
echo '{"data": {"id":"default", x: 42}}' | http PUT http://0.0.0.0:8888/v1/buckets/default -a test:test
HTTP/1.1 400 Bad Request
Access-Control-Expose-Headers: Alert, Retry-After, Backoff, Content-Length
Content-Length: 207
Content-Type: application/json; charset=UTF-8
Date: Sun, 29 May 2016 16:07:27 GMT
Server: waitress

{
    "code": 400, 
    "details": [
        {
            "description": "Record id does not match existing record", 
            "location": "body", 
            "name": "id"
        }
    ], 
    "errno": 107, 
    "error": "Invalid parameters", 
    "message": "Record id does not match existing record"
}
```

To be noted this works fine if we pass the actual uuid of the default bucket. Feel free to discard and file a server issue if this not an intended behavior.

feedback=? @leplatrem @natim @almet


Natim:
The id of the default bucket is not "default" that's the reason why you've got this error message.

Natim:
r+

n1k0:
Released in v0.8.1

leplatrem:
Let's keep like that for now, but I don't like the `if` in the clients code for the `default` bucket.

IMO it should be transparent to use. I will see if I can inspect and find a fix in the server plugin instead..

n1k0:
> IMO it should be transparent to use. I will see if I can inspect and find a fix in the server plugin instead..

I should confess I was expecting something like this but wasn't sure as *default* is an alias and not an actual *id*. I'd be okay for the server to accept a matching alias though.

PR #106, opened 2016-06-17
Add hint when server is readonly (fixes #56)

### LINKED ISSUE #56: Add checks for the readonly server setting, raise for transactional ops when true
When the server is configured as readonly, we should raise early when the user attempts at performing any transactional operation, with a user-friendly error message.

Maybe getting the information from the HTTP response is good enough? Tagging as question.

leplatrem:
@n1k0 r?


n1k0:
r+

PR #110, opened 2016-06-27
Remove quotes from last_modified attribute in listRecords()

### LINKED ISSUE #109: When last_modified is passed as ETag for safety, sent headers are wrong
Using `listRecords()` the obtained `last_modified` for the collection is quoted.

If I pass the obtained value in the `last_modified` option on other methods, the sent header is quoted twice (`'""12345""'`)

See https://github.com/Kinto/kinto-http.js/blob/v1.0.0/src/requests.js#L17

> Note: another approach could be to unquote the last_modified value in the `listRecords()` result

leplatrem:
Fixes #109

### DIFF
--- a/README.md
+++ b/README.md
@@ -999,7 +999,7 @@ Sample result:
 
 ```js
 {
-  last_modified: "\"1456183930780\"",
+  last_modified: "1456183930780",
   next: <Function>,
   data: [
     {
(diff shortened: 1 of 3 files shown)


leplatrem:
@n1k0 r?

n1k0:
r+

PR #111, opened 2016-06-27
Bump v2.0.0

### LINKED ISSUE #56: Add checks for the readonly server setting, raise for transactional ops when true

### LINKED ISSUE #77: Not able to use kinto-client from a script tag.
I tried following the documentation and loaded the `kinto-client` package from a script tag, but it doesn't seem to be working.

I also built the project from latest master and checked with the minified and unminified version, and I still have the same symptoms.

Here is a capture of the console:

### LINKED ISSUE #109: When last_modified is passed as ETag for safety, sent headers are wrong

leplatrem:
```
**Breaking changes**

- Remove quotes from `last_modified` attribute in listRecords() result (fixes #109) (#110)
- Expose `KintoClient()` constructor in browserify builds (fixes #77) (#105)

**New features**

- Add hint when server is readonly (fixes #56) (#106)
```

@n1k0 r?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kinto-http",
-  "version": "1.0.0",
+  "version": "2.0.0",
   "description": "JavaScript HTTP client for the Kinto API.",
   "main": "lib/index.js",
   "scripts": {


n1k0:
r+, can you do the release next?

leplatrem:
@n1k0 

`npm ERR! you do not have permission to publish "kinto-http"`

n1k0:
@leplatrem added you to the list of npm owners :)

PR #114, opened 2016-07-08
Selection a collection documentation — Fixes #113

### LINKED ISSUE #113: client.bucket("blog").collection("posts") is not a promise anymore
I see an error in the README.md documentation at https://github.com/kinto/kinto-http.js#selecting-a-collection

It tells us you can select a collection with `client.bucket("blog").collection("posts")` and that it returns a promise as you would be able to use `client.bucket("blog").collection("posts").then(result => …)`.

But I think it is not used that way anymore despite the documentation. Am I wrong?

Natim:
r? @ticabri

### DIFF
--- a/README.md
+++ b/README.md
@@ -739,8 +739,7 @@ Sample result:
 ### Selecting a collection
 
 ```js
-client.bucket("blog").collection("posts")
-  .then(result => ...);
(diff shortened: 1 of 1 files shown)


Natim:
r+ in #113

PR #116, opened 2016-07-15
Generalize list signatures

### LINKED ISSUE #46: Same signature for all lists ?
Currently, only the list of records can be sorted, filtered, paginated.

Since the bucket, collection, and group list share the same capabilities (it's implemented [using the same resource](http://kinto.readthedocs.org/en/latest/concepts.html#technical-architecture)), we could also apply the same behaviour to every list method :)

(not a priority though).

leplatrem:
Fixes #46

* [x] Generalize code
* [x] Leverage code in list of records
* [x]  Leverage code in list of buckets
* [x]  Leverage code in list of groups
* [x]  Leverage code in list of collections
* [x] Wait https://github.com/Kinto/kinto/pull/718
* [x] Update documentation

### DIFF
--- a/src/base.js
+++ b/src/base.js
@@ -369,7 +369,7 @@ export default class KintoClientBase {
     return raw ? promise : promise.then(({json}) => json);
   }
 
-  paginatedList(path, params, options) {
+  paginatedList(path, params, options={}) {
     const { sort, filters, limit, pages, since } = {
       sort: "-last_modified",
       ...params
--- a/src/base.js
+++ b/src/base.js
@@ -1,6 +1,6 @@
 "use strict";
 
-import { partition, pMap, omit, support, nobatch, toDataBody } from "./utils";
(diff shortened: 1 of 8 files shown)


test/integration_test.js
```diff
@@ -220,18 +220,72 @@ describe("Integration tests", function() {
       });
     });
 
-    describe("#listBuckets", () => {
+    describe.only("#listBuckets", () => {
```
leplatrem: leftover

src/base.js
```diff
@@ -366,6 +366,70 @@ export default class KintoClientBase {
     return options.raw ? promise : promise.then(({json}) => json);
   }
 
+  paginatedList(path, listOptions, httpOptions) {
```
leplatrem: * listOptions -> params
* httpOptions -> options

leplatrem:
@n1k0 final r+ ?

n1k0:
r+ \o/

PR #120, opened 2016-07-19
Fix #119: Ensure a multipart Content-Type header is used when sending attachments.

### LINKED ISSUE #119: No content-type provided when sending multipart request body
Post #115 bug:

n1k0:
Refs #119. This is ugly, but I couldn't find any better generic way to check for a multipart request body :disappointed:  

Tested the patch on kinto-admin, solves the issue.

r=? @leplatrem

### DIFF
--- a/src/http.js
+++ b/src/http.js
@@ -78,6 +78,11 @@ export default class HTTP {
     let response, status, statusText, headers, hasTimedout;
     // Ensure default request headers are always set
     options.headers = {...HTTP.DEFAULT_REQUEST_HEADERS, ...options.headers};
+    // If a multipart body is provided, remove any custom Content-Type header as
+    // the fetch() implementation will add the correct one for us.
+    if (options.body && typeof options.body.append === "function") {
+      delete options.headers["Content-Type"];
+    }
     options.mode = this.requestMode;
     return new Promise((resolve, reject) => {
       const _timeoutId = setTimeout(() => {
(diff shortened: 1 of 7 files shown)


test/kinto.ini
```diff
@@ -20,6 +20,8 @@ kinto.attachment.folder = {bucket_id}/{collection_id}
 kinto.attachment.keep_old_files = true
 kinto.attachment.base_path = /tmp
 
+kinto.experimental_permissions_endpoint = true
+
```
leplatrem: nit: leftover?

src/http.js
```diff
@@ -78,6 +78,11 @@ export default class HTTP {
     let response, status, statusText, headers, hasTimedout;
     // Ensure default request headers are always set
     options.headers = {...HTTP.DEFAULT_REQUEST_HEADERS, ...options.headers};
+    // If a multipart body is provided, remove any custom Content-Type header as
+    // the fetch() implementation will add the correct one for us.
+    if (options.body && typeof options.body.append === "function") {
+      delete options.headers["Content-Type"];
+    }
```
leplatrem: It is not worth than the previous `undefined` we had...

leplatrem:
r+

test/kinto.ini
n1k0: Woops, good catch

PR #123, opened 2016-07-20
Fix #122: Add support for the /permissions endpoint.

### LINKED ISSUE #122: Add a method to list all permissions
Kinto docs http://kinto.readthedocs.io/en/stable/api/1.x/permissions.html#list-every-permissions

Suggested API:

```js
client.listPermissions();
```

Note that it should check for the `permissions_endpoint` capability.

n1k0:
refs #122, r=? @leplatrem 

## Listing all resource permissions

If the [`permissions_endpoint` capability](http://kinto.readthedocs.io/en/stable/api/1.x/permissions.html#list-every-permissions) is installed on the server, you can retrieve the list of all permissions set for the authenticated user using the `listPermissions()` method:

```js
client.listPermissions([options])
  .then(result => ...);
```

Sample result:

```js
{
    "data": [
        {
            "bucket_id": "mybucket",
            "id": "mybucket",
            "permissions": [
                "write",
                "read",
                "group:create",
                "collection:create"
            ],
            "resource_name": "bucket",
            "uri": "/buckets/mybucket"
        },
        ...
    ]
}
```

### DIFF
--- a/test/api_test.js
+++ b/test/api_test.js
@@ -700,33 +700,44 @@ describe("KintoClient", () => {
   describe("#listPermissions()", () => {
     const data = [{id: "a"}, {id: "b"}];
 
-    beforeEach(() => {
-      api.serverInfo = {capabilities: {"permissions_endpoint": {}}};
-      sandbox.stub(api, "execute").returns(Promise.resolve({data}));
-    });
+    describe("Capability available", () => {
+      beforeEach(() => {
+        api.serverInfo = {capabilities: {"permissions_endpoint": {}}};
+        sandbox.stub(api, "execute").returns(Promise.resolve({data}));
+      });
 
-    it("should execute expected request", () => {
-      api.listPermissions()
-        .then(() => {
-          sinon.assert.calledWithMatch(api.execute, {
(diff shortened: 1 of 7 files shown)


test/api_test.js
```diff
@@ -540,6 +540,40 @@ describe("KintoClient", () => {
     });
   });
 
+  /** @test {KintoClient#listPermissions} */
+  describe("#listPermissions()", () => {
+    const data = [{id: "a"}, {id: "b"}];
+
```
leplatrem: I don't if we did for other stuff, but it may be worth adding a test to check that it fails nicely when capability is not present ?

leplatrem:
r+wc

test/api_test.js
n1k0: Dunno as `capable` is pretty well tested already... ok why not, never too safe I guess :)

n1k0: Addressed in https://github.com/Kinto/kinto-http.js/pull/123/commits/75d80014a874c0af39b5688731946885aff2a3f3

PR #126, opened 2016-08-19
Fix #124: Fake headers object in batch list operations.

### LINKED ISSUE #124: Do not try consumming HTTP response headers in batch requests
When we generalized list params & pagination in #116, we introduced a bug where we attempt at consuming response headers while none are available in the fake response object in batch. 

/cc @leplatrem

n1k0:
r=? @leplatrem


leplatrem:
r+ !

PR #130, opened 2016-08-25
Allow simple checking for the existence of a next page.

n1k0:
This one will probably trigger interesting discussions :)

While working on implementing support for list pagination in the kinto admin, I realized the `next()` mechanism for paginating results was flawed from a developer usability standpoint; to check if there's a next page (eg. to render a *Next page* button), one has to actually execute the function, which actually performs the HTTP request. There's obviously a valid use case where you simply want to know if there **is** a next page or not.

I have to fallback to the good old `Function | null` pattern to know if a next page is retrievable; I'm not super easy with it, but in this specific case I'm finding it simple yet quite efficient. If you have better patterns in mind to address this, feel free to mention them in the comments.

Feedback=? @leplatrem @Natim @glasserc 

**Update:** Alternative PR in #131

### DIFF
--- a/README.md
+++ b/README.md
@@ -1392,7 +1392,7 @@ client.bucket("blog").collection("posts")
   .then(({data, next}) => {
 ```
 
-To retrieve the next page of results, you can check for the `next` property attached to the result object obtained. If a next page is available, `next` is a function, while if pagination is exhausted, `next` is `null`:
+To retrieve the next page of results, you can check for the `next` property attached to the result object obtained. If a next page is available, `next` is a function you can call to retrieve the next page of results, and becomes a `null` when pagination is exhausted:
 
 ```js
 let getNextPage;
@@ -1414,6 +1414,8 @@ if (getNextPage) {
       console.log("Page 2", data);
       getNextPage = next; // etc...
     });
+} else {
+  console.log("No more pages.")
 }
 ```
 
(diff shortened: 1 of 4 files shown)


n1k0:
Ah, and as it clearly breaks the public API, if this lands it requires bumping a major :/

n1k0:
We could alternatively have `next()` itself returning either a function or a `null`:

```js
let getNextPage;

client.bucket("blog").collection("posts")
  .listRecords({limit: 20})
  .then(({data, next}) => {
    console.log("Page 1", data);
    getNextPage = next();
  });
```

Later down the flow:

```js
if (getNextPage) {
  getNextPage()
    .then(({data, next}) => {
      console.log("Page 2", data);
      getNextPage = next(); // etc...
    });
} else {
  console.log("No more pages.")
}
```

this would still break the oublic API though may be slightly better. What do you think?

Natim:
Cannot we add a new variable hasNextPage set to true when next is a function and false when no more pages?

n1k0:
> Cannot we add a new variable hasNextPage set to true when next is a function and false when no more pages?

That's what I went with at first but found it redundant with just checking the `next` type. I may get back to this if everybody feels better with this approach.

n1k0:
Alternative proposal in #131.

n1k0:
Discarded in favor of #131.

PR #131, opened 2016-08-25
Expose a hasNextPage property to the list results object.

n1k0:
Alternative proposal to #130.

Feedback=? @Natim @leplatrem @glasserc


n1k0:
Landing this after discussing this with @leplatrem.

PR #133, opened 2016-08-27
fix #127 - Expose the endpoint function to client programs

### LINKED ISSUE #127: Expose the endpoint function to client programs
`endpoint()` is currently not available to client programs.
If they want to make raw queries to the kinto server, using kinto.js, they have no way to use  ̀endpoint()` in 

```
var api = collection.api.bucket(collection.bucket).collection(collection.name)
api.client.execute({
 path: endpoint("collection", this.bucket.name, this.name),
 method: "HEAD",
 raw: true,
 headers,
})
```

They also need to add authentication, but it's the subject of another issue.

elelay:
I've created a test, but the endpoint function can't be stubbed AFAICT, so I've compared one output.
Coverage should be good.
I've not kept the `(name, ...args)` arguments, because the generated code is silly: deconstructs and constructs again the `arguments`.

### DIFF
--- a/src/base.js
+++ b/src/base.js
@@ -524,4 +524,15 @@ export default class KintoClientBase {
     const path = endpoint("bucket");
     return this.execute(requests.deleteRequest(path, reqOptions));
   }
+
+  /**
+   * Retrieves a server enpoint by its name.
+   *
+   * @param  {String}    name The endpoint name.
+   * @param  {...string} args The endpoint parameters.
+   * @return {String}
+   */
+  static endpoint() {
+    return endpoint.apply(undefined, arguments);
+  }
 }
--- a/test/api_test.js
+++ b/test/api_test.js
@@ -10,6 +10,8 @@ import { SUPPORTED_PROTOCOL_VERSION as SPV } from "../src/base";
 import * as requests from "../src/requests";
 import Bucket from "../src/bucket";
 
+import endpoint from "../src/endpoint";
+
 chai.use(chaiAsPromised);
 chai.should();
 chai.config.includeStack = true;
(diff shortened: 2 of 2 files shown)


Natim:
The plan is to make sure people who are using kinto-http.js do not have to rely on this low level api.

It might be complex to achieve with plugins creating new endpoint but in that case we need to think of a better way to help people create URLs.

Natim:
I think we all agree that the path forward would be to implement this API:

    var collection_uri = tickets.api.bucket(tickets.bucket).collection(tickets.name).endpoint();
    var collection_records_uri = tickets.api.bucket(tickets.bucket).collection(tickets.name).endpoint('record');
    var collection_history_uri = tickets.api.bucket(tickets.bucket).collection(tickets.name).endpoint('history');

elelay:
And also this (not restricting what enpoint can be computed from collection api, bucket api, etc.) ?
```javascript
var buckets_uri_from_collection = tickets.api.bucket(tickets.bucket).collection(tickets.name).endpoint('bucket');
```


This could be achieved by passing an object with all parameters to the endpoint function and it uses whatever key it needs:

**endpoint.js**:
```javascript
/**
 * Endpoints templates.
 * @type {Object}
 */
const ENDPOINTS = {
  bucket: (params) =>
    "/buckets" + (params.bucket ? `/${params.bucket}` : ""),
  collection: (params) =>
    `${ENDPOINTS.bucket(params)}/collections` + (params.collection ? `/${params.collection}` : ""),
  ...
};

/**
 * Retrieves a server enpoint by its name.
 *
 * @private
 * @param  {String}    name The endpoint name.
 * @param  {Object} params The endpoint parameters.
 * @return {String}
 */
export default function endpoint(name, params) {
  return ENDPOINTS[name](params);
}
```

**collection.js**:
```javascript
export default class Collection {
    endpoint(endpointName, params) {
       return endpoint({
           ...params,
           bucket: this.bucket.name,
           collection: this.name
       });
  }
}
```

elelay:
The code in my PR is wrong!
Closing it.

PR #134, opened 2016-08-28
Join multiple same-key parameters with a comma. Fixes #129.

### LINKED ISSUE #129: Broken filtering (+ broken link in the docs)
There is a broken link in the docs. At https://www.npmjs.com/package/kinto-http#generic-options-for-list-operations the "what's supported" link gives 404 because it tries to access http://kinto.readthedocs.io/en/stable/core/api/resource.html#filtering.

There is a  http://kinto.readthedocs.io/en/stable/core/resource.html page (without /api/), but it doesn't have any #filtering element.

***
I was looking for filtering documentation because when I try to
`collection.listRecords({ filters: { id: ['id1', 'id2'] } })`
I get empty results, but when I try 
`collection.listRecords({ filters: { id: ['id1']  } })` 
and 
`collection.listRecords({ filters: { id: ['id2'] } })`
I get 1 result both times. In this case, shouldn't the first snippet return both records? 

I tried this on a different collection with a different property (to exclude some bug with 'id's), and I get same buggy result.

doplumi:


src/utils.js
```diff
@@ -89,7 +89,8 @@ export function qsify(obj) {
   return Object.keys(stripped).map((k) => {
     const ks = encode(k) + "=";
     if (Array.isArray(stripped[k])) {
-      return stripped[k].map((v) => ks + encode(v)).join(sep);
+      const arraySep = ","; 
+      return ks + stripped[k].map((v) => encode(v)).join(arraySep);
```
n1k0: nit: no need for this intermediary `arraySep` const imho, and `join` is explicit enough

n1k0:
LGTM, r+ with the link to the kinto docs fixed as well as discussed in #129

src/utils.js
doplumi: I'm OK with that. Also, the `sep` constant declared at the top of the function became unnecessary with the change at line 92. I'm gonna remove both.

doplumi:
Done. Also fixed all of other broken `../core/api/..` links.

n1k0:
Thanks! I'll release a new version soon.

n1k0:
Released in [v2.4.1](https://github.com/Kinto/kinto-http.js/releases/tag/v2.4.1).

PR #142, opened 2016-09-16
implement #138 since option for collection.getTotalRecords

elelay:
getTotalRecords(options) has been released already, so this is a breaking API change but not many have jumped on it I hope...

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -77,16 +77,16 @@ export default class Collection {
   /**
    * Retrieves the total number of records in this collection.
    *
-   * @param  {String} [since]           The since ETag param (optional).
    * @param  {Object} [options={}]      The options object.
    * @param  {Object} [options.headers] The headers object option.
+   * @param  {String} [options.since]   The since ETag option.
    * @return {Promise<Number, Error>}
    */
-  getTotalRecords(since, options={}) {
+  getTotalRecords(options={}) {
     const { headers } = this._collOptions(options);
     let path = endpoint("record", this.bucket.name, this.name);
-    if(since){
-      path += "?since="+since;
+    if(options.since){
+      path += "?since="+options.since;
     }
     return this.client.execute({
       method: "HEAD",
--- a/src/collection.js
+++ b/src/collection.js
@@ -1,6 +1,6 @@
 import { v4 as uuid } from "uuid";
 
-import { capable, toDataBody, isObject } from "./utils";
+import { capable, toDataBody, isObject, qsify } from "./utils";
 import * as requests from "./requests";
 import endpoint from "./endpoint";
 
@@ -83,14 +83,15 @@ export default class Collection {
    * @return {Promise<Number, Error>}
    */
   getTotalRecords(options={}) {
-    const { headers } = this._collOptions(options);
+    const { since, ...otherOptions } = options;
+    const { headers } = this._collOptions(otherOptions);
     let path = endpoint("record", this.bucket.name, this.name);
-    if(options.since){
-      path += "?since="+options.since;
+    if (since) {
+      path += "?" + qsify({since});
     }
     return this.client.execute({
       method: "HEAD",
-      path: path,
+      path,
       headers
     }, {raw: true})
       .then(({headers}) => parseInt(headers.get("Total-Records"), 10));
--- a/src/collection.js
+++ b/src/collection.js
@@ -77,15 +77,20 @@ export default class Collection {
   /**
    * Retrieves the total number of records in this collection.
    *
+   * @param  {String} [since]           The since ETag param (optional).
    * @param  {Object} [options={}]      The options object.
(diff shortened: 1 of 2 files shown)


src/collection.js
```diff
@@ -77,15 +77,20 @@ export default class Collection {
   /**
    * Retrieves the total number of records in this collection.
    *
+   * @param  {String} [since]           The since ETag param (optional).
    * @param  {Object} [options={}]      The options object.
    * @param  {Object} [options.headers] The headers object option.
    * @return {Promise<Number, Error>}
    */
-  getTotalRecords(options={}) {
+  getTotalRecords(since, options={}) {
```
leplatrem: I think we should follow what was done in `listRecords()` : make `since` an option. Plus, it wouldn't break the API!

test/collection_test.js
```diff
@@ -54,6 +54,18 @@ describe("Collection", () => {
...
+    it("should pass the since parameter", () => {
+      sandbox.stub(client, "execute").returns(Promise.resolve());
+
+      getBlogPostsCollection().getTotalRecords("ETAG");
+
+      sinon.assert.calledWithMatch(client.execute, {
+        method: "HEAD",
+        path: "/buckets/blog/collections/posts/records?since=ETAG",
+      }, {raw: true});
+    });
```
leplatrem: 👍

src/collection.js
```diff
@@ -79,13 +79,18 @@ export default class Collection {
...
    * @param  {Object} [options={}]      The options object.
    * @param  {Object} [options.headers] The headers object option.
+   * @param  {String} [options.since]   The since ETag option.
    * @return {Promise<Number, Error>}
    */
   getTotalRecords(options={}) {
     const { headers } = this._collOptions(options);
+    let path = endpoint("record", this.bucket.name, this.name);
+    if(options.since){
+      path += "?since="+options.since;
```
n1k0: It may be nice to reuse the `utils.qsify` function here.

Also nit: space around operand

src/collection.js
```diff
@@ -79,13 +79,18 @@ export default class Collection {
    *
    * @param  {Object} [options={}]      The options object.
    * @param  {Object} [options.headers] The headers object option.
+   * @param  {String} [options.since]   The since ETag option.
    * @return {Promise<Number, Error>}
    */
   getTotalRecords(options={}) {
     const { headers } = this._collOptions(options);
+    let path = endpoint("record", this.bucket.name, this.name);
+    if(options.since){
```
n1k0: nit: space after `if` and before `{`

src/collection.js
```diff
@@ -79,13 +79,18 @@ export default class Collection {
    *
    * @param  {Object} [options={}]      The options object.
    * @param  {Object} [options.headers] The headers object option.
+   * @param  {String} [options.since]   The since ETag option.
    * @return {Promise<Number, Error>}
    */
   getTotalRecords(options={}) {
     const { headers } = this._collOptions(options);
```
n1k0: So that means `since` is passed to `_collOptions` and merged into the instance options. I don't think that it's a problem per se, but explicitely removing it may be nicer:

```js
getTotalRecords(options={}) {
  const { since, ...otherOptions } = options;
  const { headers } = this._collOptions(otherOptions);
  let path = endpoint("record", this.bucket.name, this.name);
  if (since) {
    path += "?since="+options.since;
  }
```

src/collection.js
```diff
@@ -77,15 +77,20 @@ export default class Collection {
...
+  getTotalRecords(since, options={}) {
     const { headers } = this._collOptions(options);
+    let path = endpoint("record", this.bucket.name, this.name);
+    if(since){
+      path += "?since="+since;
+    }
     return this.client.execute({
       method: "HEAD",
-      path: endpoint("record", this.bucket.name, this.name),
+      path: path,
```
n1k0: I insist on this, pretty please:

```js
return this.client.execute({
    method: "HEAD",
    path,
    headers,
  }, {raw: true})
```

elelay: I've used path directly. Is there something else to do?

src/collection.js
elelay: actually, the `const { since, ...otherOptions } = options` instruction breaks the firefox build:
https://travis-ci.org/Kinto/kinto-http.js/jobs/163077307#L959

Maybe this issue: https://github.com/substack/module-deps/issues/93

n1k0: Sigh. Oh well.

src/collection.js
n1k0: That's basically the difference between:

```js
return this.client.execute({
    method: "HEAD",
    path: path,
    headers,
  }, {raw: true})
```

and

```js
return this.client.execute({
    method: "HEAD",
    path,
    headers,
  }, {raw: true})
```

PR #148, opened 2016-11-03
Fix #147: safe creation of attachment

### LINKED ISSUE #147: Allow using the safe option when creating a record with addAttachment
Refs https://github.com/Kinto/kinto-admin/pull/305

leplatrem:
@n1k0 r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ node_js:
 python:
 - "3.5"
 env:
-- SERVER=4.3.0 ACTION=test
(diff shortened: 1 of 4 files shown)


n1k0:
Looks good (thanks!) though these test failures are worrying (they don't seem to be related to your patch though; latest python package versions?)

leplatrem:
Tests fail because kinto-attachment is not compatible with current Kinto master. #willdo

PR #155, opened 2017-01-12
Fix #34: honour retry-after

### LINKED ISSUE #34: Handle Retry-After on HTTP 503

leplatrem:
In some unexpected situations, the server may fail to process the request and ask the client to retry it.

In this PR, the request is retried once by default after the time specified by the server. 

From the usage perspective of the library (e.g. kinto.js sync), everything remains the same, it's just as if the server would have taken more time to respond.


test/http_test.js
```diff
@@ -259,21 +259,48 @@ describe("HTTP class", () => {
...
+            .should.eventually.become(success);
+        }).timeout(1100);
+
+        it("should error when retries are exhausted", () => {
+          fetch.onCall(0).returns(fakeServerResponse(503, {}, {"Retry-After": "1"}));
+          fetch.onCall(1).returns(fakeServerResponse(503, {}, {"Retry-After": "1"}));
+          fetch.onCall(2).returns(fakeServerResponse(503, {}, {"Retry-After": "1"}));
+          return http.request("/", {retry: 2})
+            .should.eventually.be.rejectedWith(Error, /Error: HTTP 503/);
+        }).timeout(2100);
```
leplatrem: 😿

src/http.js
```diff
@@ -72,9 +72,10 @@ export default class HTTP {
    * @param  {String} url               The URL.
    * @param  {Object} [options={}]      The fetch() options object.
    * @param  {Object} [options.headers] The request headers object (default: {})
+   * @param  {Object} [options.retry]   Number of retries (default: 1)
    * @return {Promise}
    */
-  request(url, options={headers:{}}) {
+  request(url, options={headers:{}, retry: 1}) {
```
n1k0: Shouldn't we default `retry` to `0`, so by default there's no retry attempt on failure like previously?

test/http_test.js
n1k0: You can simulate these using sinon's fake timer I think http://sinonjs.org/docs/#clock

n1k0: Oh wait, is it a timeout set for the `it` statement? Why do we need this?

src/http.js
leplatrem: Yes, why not. But in most cases, retrying once will fix the problem so I thought it was a good idea... No problem to set to 0 if you feel sure

test/http_test.js
leplatrem: This is very ugly indeed, it is just to control mocha timeout for this specific test.

I didn't succeed to use fake timers, because I had to `clock.tick()` once the errored response was received. I tried to tick from a `events.on('retry-after')` callback but it had no effect.

IMO this is really bad because it really slows down the tests suite execution!!

src/http.js
n1k0: I usually prefer calling specific behaviors explicitly, but in this case we could probably live with a songle retry performed by default. Your call.

leplatrem: I followed the python client behaviour: no retry by default.

n1k0:
I'm gonna clone the branch and experiment with fake timers to see if I can come with another approach in tests :)

src/http.js
n1k0: According to https://github.com/Kinto/kinto-http.js/pull/155/commits/915c6b152b402e9032553324f20342ca2260b117 should it be 0 by default now?

PR #156, opened 2017-01-17
Fix Gecko unknown symbol

leplatrem:
screenshot from 2017-01-17 15-20-40


Natim:
If it is generated from kinto-http.js shouldn't we add something in the generation script?

leplatrem:
Thanks :heart

PR #158, opened 2017-02-09
Get rid of polyfill and upgrade dependencies

leplatrem:
I don't think we should require polyfills in the code. (see https://github.com/Kinto/kinto.js/issues/650)

Users should use a polyfill service if needed, like https://cdn.polyfill.io/v2/docs/features/

This PR:
* keeps isomorphic-fetch for tests with mocha only
* upgrade js dependencies
* upgrade server version for tests


n1k0:
https://twitter.com/n1k0/status/829641097605431297

leplatrem:
🤣 

With a broken knee!

PR #159, opened 2017-02-17
Reducing library size

screendriver:
Same as https://github.com/Kinto/kinto.js/pull/652


screendriver:
I'm stuck here a little bit. Not because of the technical implementation but I'm thinking about what should be included in the build and what not. For instance should there be a polyfill for Promises and the fetch API?

For **me** a library should be referenced in my project as a npm dependency and **I** am responsible that everything should be there at runtime. So **my** project should ensure that everything exists. But I am using a "professional" build system, transpilation, polyfilling and so on. There are also developers out there that reference `kinto` as a good old `<script>` tag over a CDN. For this people it makes no sense to exclude everything from the build. For these developers **everything** should be included in the build to ensure that it runs automatically in all modern and old browsers.

So I don't know how to resolve this smartest. Maybe there should be different builds:

- one for developers that consume the library as a `npm` dependency. Here is the question: already transpiled but without any polyfills? Only the raw sources? If yes then the consuming project has to tell their Babel transpiler that this project should also be transpiled (I think that is a not so good idea). If already transpiled then tree shaking will not be possible
- one for developers that consume the library in a good old `<script>` tag. Then it must include everything. Transpiled to ES5, include every polyfill and so on
- one for the Firefox browser (everything transpiled, but without the polyfill)

Also I'm thinking about to switch to [rollup.js](http://rollupjs.org). Not because webpack produces bad bundles. It's because `rollup.js` is made for libraries that don't use assets like `*.png`, `*.css` and so on. It "only" bundles JavaScript but extremely good. This is just a side note and has nothing to do with my thoughts 😉

screendriver:
Oh and by the way: I'm thinking about to add [jsnext:main](https://github.com/rollup/rollup#can-i-distribute-my-package-as-an-es6-module) to the `package.json`

At the moment it's a little bit complicated out there because **every** library has the same "issue" at the moment like we have at the moment.

screendriver:
Sorry for the delay. I was ill and very busy in the last time.

I believe I'm done. I switched from webpack to rollup.js and added `module` / `jsnext:main` to the `package.json`. 

- `kinto-http.es.js` is meant for people with a build system like webpack, rollup,js and so forth. It does **not** include any bundled libraries. Only transpiled source code in ES2015 module syntax.
- `kinto-http.cjs.js` the same like `kinto-http.es.js` but in CommonJS.
- `kinto-http.noshim.js` the same like `kinto-http.es.js` but in UMD (it would be the best to rename it to `kinto-http.umd.js`)
- `kinto-http.min.js` everything included. All libraries, minified and in UMD
- `moz-kinto-http-client.js` should work as before (but I am not sure)

Also I added [unfetch](https://github.com/developit/unfetch). A 500 byte polyfill for the `fetch` API. I am using it as a **ponyfill** here because I don't want to collide with other polyfills that could be potentially be used in the consuming project.

I hope I don't forgot anything.

screendriver:
PS: I'm a little bit torn about webpack vs. rollup.js. I don't see any **real** difference 😳 I'm using rollup here because I wanted to try something new for me 🙄

dstaley:
I'm pretty excited to say that the new Rollup/TypeScript version of the library is now 60% smaller!

v4.7.3 is 67.8 kB minified, and 14.6 kB gzipped. 5.0.0-alpha.2 is 27 kB minified, and 7.3 kB gzipped.

Screen Shot 2019-09-19 at 6 00 05 PM

leplatrem:
Awesome!

dstaley:
As of `v5.0.0`, the library is now just 7.3 kB gzipped and minified! Gonna go ahead and close this.

PR #161, opened 2017-02-24
Add support for a 'at' collection listRecords filter.

n1k0:
This adds a new `at` option to `Collection#listRecords` to retrieve a snapshot of the collection records at a given timestamp in the past.

- [x] Compute collection list of records state at a given timestamp back in time
- [x] Ensure paginating long lists of changes
- [x] Update documentation
- [ ] Fix travis build setup

f=? @leplatrem

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -314,6 +314,7 @@ export default class Collection {
    * @param  {Object}   [options.headers]               The headers object option.
    * @param  {Object}   [options.filters=[]]            The filters object.
    * @param  {String}   [options.sort="-last_modified"] The sort field.
+   * @param  {String}   [options.at]                    The timestamp to get a snapshot at.
    * @param  {String}   [options.limit=null]            The limit field.
    * @param  {String}   [options.pages=1]               The number of result pages to aggregate.
    * @param  {Number}   [options.since=null]            Only retrieve records modified since the provided timestamp.
@@ -322,7 +323,41 @@ export default class Collection {
   listRecords(options={}) {
     const path = endpoint("record", this.bucket.name, this.name);
     const reqOptions = this._collOptions(options);
-    return this.client.paginatedList(path, options, reqOptions);
+    if (options.hasOwnProperty("at")) {
+      return this._getSnapshot(options.at);
+    } else {
+      return this.client.paginatedList(path, options, reqOptions);
+    }
+  }
+
+  _getSnapshot(at) {
+    const seenIds = new Set();
+    let snapshot = [];
+    if (!Number.isInteger(at) || at <= 0) {
+      throw new Error("Invalid argument, expected a positive integer.");
+    }
+    const before = String(at);
+    return this.bucket.listHistory({
+      sort: "-target.data.last_modified",
+      filters: {
+        resource_name: "record",
+        collection_id: this.name,
+        "max_target.data.last_modified": before,
+      }
+    })
+      .then(({data: changes}) => {
+        for (const change of changes) {
+          const {data: record} = change.target;
+          if (record.deleted) {
+            seenIds.add(record.id);
+            snapshot = snapshot.filter(r => r.id !== record.id);
+          } else if (!seenIds.has(record.id)) {
+            seenIds.add(record.id);
+            snapshot = [record, ...snapshot];
+          }
+        }
+        return snapshot.sort((a, b) => a.last_modified < b.last_modified ? 1 : -1);
+      });
   }
 
   /**
--- a/src/collection.js
+++ b/src/collection.js
@@ -361,7 +361,7 @@ export default class Collection {
           }
         }
         return {
-          last_modified: Math.max.apply(null, snapshot.map(r => r.last_modified)),
+          last_modified: String(at),
           data: snapshot.sort((a, b) => b.last_modified - a.last_modified),
           next: () => { throw new Error("Snapshots don't support pagination"); },
           hasNextPage: false,
--- a/src/collection.js
+++ b/src/collection.js
@@ -331,32 +331,29 @@ export default class Collection {
   }
 
   _getSnapshot(at) {
-    const seenIds = new Set();
-    let snapshot = [];
     if (!Number.isInteger(at) || at <= 0) {
       throw new Error("Invalid argument, expected a positive integer.");
     }
-    const before = String(at);
     return this.bucket.listHistory({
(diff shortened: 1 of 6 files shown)


n1k0:
TGIF



computing

n1k0:
There's a bug in Kinto history that prevent paginating list of changes it seems: https://github.com/Kinto/kinto/issues/1114

src/collection.js
```diff
@@ -322,7 +323,38 @@ export default class Collection {
...
+        for (const {target: {data: record}} of changes) {
+          if (record.deleted) {
+            seenIds.add(record.id);
+            snapshot = snapshot.filter(r => r.id !== record.id);
+          } else if (!seenIds.has(record.id)) {
+            seenIds.add(record.id);
+            snapshot.push(record);
+          }
+        }
+        return snapshot.sort((a, b) => b.last_modified - a.last_modified);
```
leplatrem: nit: should we respect the `sort` option provided in `listRecords()`?

test/integration_test.js
```diff
@@ -1430,6 +1430,56 @@ describe("Integration tests", function() {
               });
             });
 
+            describe("'at' retrieves a snapshot at a given timestam", () => {
```
leplatrem: typo: *timestamp*

test/integration_test.js
```diff
@@ -1430,6 +1430,56 @@ describe("Integration tests", function() {
...
+              let rec1, rec2, rec3;
+
+              beforeEach(() => {
+                return coll.listRecords()
```
leplatrem: nit: why do we list first?

test/integration_test.js
```diff
@@ -1430,6 +1430,56 @@ describe("Integration tests", function() {
...
+                  .then(({data}) => {
+                    rec2 = data;
+                    return coll.createRecord({n: 3});
+                  })
+                  .then(({data}) => rec3 = data);
+              });
+
+              it("should handle creations", () => {
+                return coll.listRecords({at: rec1.last_modified})
+                  .should.eventually.become([
```
leplatrem: I think it should return the same structure as `_paginatedList` : 
https://github.com/Kinto/kinto-http.js/blob/master/src/base.js#L410-L415

src/collection.js
```diff
@@ -322,7 +323,38 @@ export default class Collection {
...
+    } else {
+      return this.client.paginatedList(path, options, reqOptions);
+    }
+  }
+
+  _getSnapshot(at) {
+    if (!Number.isInteger(at) || at <= 0) {
+      throw new Error("Invalid argument, expected a positive integer.");
+    }
+    return this.bucket.listHistory({
```
leplatrem: We could add a XXX comment  that explains that we take the history forward, and that it would more efficient (and more chance to be complete) if we'd do it backward

n1k0:
> I find it a lot easier to handle than on the server... No?

Having a server side implementation is much more of a strategic move then a technical one imho, but as you highligted in previous comments if the kinto-admin is the only project that would benefit from it it indeed challenges the "strategy" quite a lot ;)

test/integration_test.js
n1k0: So you'd want to allow paginating a snapshot?

test/integration_test.js
n1k0: A leftover most probably, will remove.

Natim:
To be honest my call on making it a server feature was to move forward the issue about syncing between Kinto and CouchDB. Issue that was brought to us by CozyCloud.

src/collection.js
n1k0: That's a good idea, I'm just slightly afraid that would open the door to supporting all other options. I liked the opinionated simplistic approach here, but we may want to think about it further.

src/collection.js
n1k0: We could also update the code to process the history backward directly if that would prevents later issues.

test/integration_test.js
n1k0: Poke, I'd really like to know about that, and also if you have a plan in mind to achieve it simply :)

n1k0: Or would something like this be good enough?

```js
        return {
          last_modified: Math.max.apply(null, snapshot.map(r => r.las_modified)),
          data: snapshot.sort((a, b) => b.last_modified - a.last_modified),
          next: () => { throw new Error("Snapshots don't support pagination"); },
          hasNextPage: false,
          totalRecords: snapshot.length,
        };
```

n1k0:
I think I've addressed most comments in https://github.com/Kinto/kinto-http.js/pull/161/commits/31674f78d4d30ed591d897ae6d62b1c7ee606609. We could always iterate further in next PRs to improve the implementation.

New feedback requested @leplatrem @Natim

src/collection.js
```diff
@@ -322,7 +323,51 @@ export default class Collection {
...
+          if (record.deleted) {
+            seenIds.add(record.id); // ensure not reprocessing deleted entries
+            snapshot = snapshot.filter(r => r.id !== record.id);
+          } else if (!seenIds.has(record.id)) {
+            seenIds.add(record.id);
+            snapshot.push(record);
+          }
+        }
+        return {
+          last_modified: Math.max.apply(null, snapshot.map(r => r.last_modified)),
```
leplatrem: A priori, this would just be `last_modified: String(at)` ?

n1k0: Not necessarily as the `at` parameter may be higher than the latest record timestamp. I felt more accurate to rely on the actual data. Also, the collection last_modified is supposed to be a number I think.

n1k0:
Cool, now I "just" need to fix the travis build setup, while I can't access the build logs anymore (aws issue still ongoing?)

src/collection.js
leplatrem: @n1k0 I realize this would give a numeric value, whereas we want an ETag. Just use `String(at)` and you're good

n1k0: Wow I realize we expose the ETag in the list result object in kinto-http, while Kinto exposes `last_modified` as an integer in the collection attributes. This is rather confusing tbh. Oh well, will update as suggested.

n1k0: Done in https://github.com/Kinto/kinto-http.js/pull/161/commits/ce8a1c763664906ce85afb91f37e912a0488a061

n1k0:
I'm landing this.

PR #163, opened 2017-03-06
Backward history powered snapshots

n1k0:
r=? @leplatrem

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -6,6 +6,44 @@ import endpoint from "./endpoint";
 
 
 /**
+ * Computes a snapshot of a collection at a given timestamp, from a current list
+ * of records and a list of changes.
+ */
+export function computeSnapshotAt(at, records, changes) {
+  // Create a map with current records, indexed by id for convenience
+  const recordsByIds = {};
+  for (const record of records) {
+    recordsByIds[record.id] = record;
+  }
+
+  // Process the changes, which are listed backward, so we start with the most
+  // recent change up to the oldest one; delete created entries, restore deleted
+  // and updated ones
+  for (const change of changes) {
+    const {action, target: {data: entry}} = change;
+    if (action === "create") {
+      delete recordsByIds[entry.id];
+    } else {
+      recordsByIds[entry.id] = entry;
+    }
+  }
+
+  // Ensure sorting the resulting records by last_modified desc.
+  const snapshot = Object.values(recordsByIds)
+    .sort((a, b) => b.last_modified - a.last_modified);
+
+  // If we find a timestamp greater than what has been required initially,
+  // that means that we couldn't retrieve enough history data to compute
+  // a consistent snapshot.
+  const newest = Math.max.apply(null, snapshot.map(r => r.last_modified));
+  if (newest > at) {
+    throw new Error("Could not compute snapshot: not enough history data.");
+  }
+
+  return snapshot;
+}
+
+/**
  * Abstract representation of a selected collection.
  *
  */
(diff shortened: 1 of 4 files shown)


src/collection.js
```diff
@@ -337,32 +375,23 @@ export default class Collection {
...
-            snapshot = snapshot.filter(r => r.id !== record.id);
-          } else if (!seenIds.has(record.id)) {
-            seenIds.add(record.id);
-            snapshot.push(record);
-          }
-        }
+    });
+    return Promise.all([listRecords, listHistory])
+      .then(([{data: records}, {data: changes}]) => {
+        const snapshot = computeSnapshotAt(at, records, changes);
```
leplatrem: 👍

test/collection_test.js
```diff
@@ -420,3 +420,15 @@ describe("Collection", () => {
     });
   });
 });
+
+describe("computeSnapshotAt()", () => {
+  // Note: most of the tests covering this API are done in integration_test.js
+
+  it("should raise when a snapshot records timestamp exceeds requirements", () => {
```
leplatrem: nit: *exceeds period covered by changes* maybe?

test/collection_test.js
```diff
@@ -420,3 +420,15 @@ describe("Collection", () => {
...
+    const records = [{id: 1, last_modified: 43}, {id: 2, last_modified: 44}];
+    const changes = [{action: "create", target: {data: {last_modified: 40}}}];
```
leplatrem: If we want to simulate that the `history` plugin was enabled after the record `id: 1` was created, then we should have `last_modified: 44` (or 45, but above 43)

Imagine the following scenario:

- `id: 1` is created (ts: 43)
- plugin is enabled 
- `id: 2`is created (ts: 44), change entry `action: create` is tracked (ts >= 44)

test/collection_test.js
```diff
@@ -420,3 +420,15 @@ describe("Collection", () => {
...
+
+    expect(() => computeSnapshotAt(42, records, changes))
```
leplatrem: should be below oldest available change

src/collection.js
```diff
@@ -6,6 +6,44 @@ import endpoint from "./endpoint";
 
 
 /**
+ * Computes a snapshot of a collection at a given timestamp, from a current list
+ * of records and a list of changes.
+ */
+export function computeSnapshotAt(at, records, changes) {
```
leplatrem: meganit: I would have put `at` as last argument :)

n1k0:
After more checking, it looks like history deletion entries only have a tombstone as a target:



That makes this patch unable to deal with restoring deleted entries I think. Thoughts?

n1k0:
Updated implementation in https://github.com/Kinto/kinto-http.js/pull/163/commits/71924f60abc90307765060ad487be408e4b99ebe. I think we should carefully review the tests, possibly add more to ensure we're not missing any edge case.

n1k0:
Ok, during my latest investigation I've stumbled upon a limitation in ongoing implementation where updated records weren't properly picked by the backward algorythm. I've pushed new tests on master in https://github.com/Kinto/kinto-http.js/commit/585d5518f670469c112944ffcc5c2e21fe356f3d to ensure it was working with the previous one, so they're failing now with this branch.

We'll have to fix this. TOMORROW. 'night.

n1k0:
Funnily enough we don't have unit and integration tests running anymore on Travis:

n1k0:
Tests enabled again. Kinto server setup now needs proper fixing

n1k0:
Ok at least we now have our test suite running again, and we have the two errors we should fix with our implementation (plus another one not related to this patch):

n1k0:
Note that I've created a separate PR to track the other failing test issue in https://github.com/Kinto/kinto-http.js/pull/164

leplatrem:
Can't be achieved with the current history API since history entries contain the current (ie. new) version of the object during the operation.

PR #165, opened 2017-03-10
Prettierify code.

n1k0:
[Prettier](https://github.com/prettier/prettier) is pretty. Let's use it.

### DIFF
--- a/src/bucket.js
+++ b/src/bucket.js
@@ -3,7 +3,6 @@ import Collection from "./collection";
 import * as requests from "./requests";
 import endpoint from "./endpoint";
 
-
 /**
  * Abstract representation of a selected bucket.
  *
@@ -18,7 +17,7 @@ export default class Bucket {
    * @param  {Object}      [options.headers] The headers object option.
    * @param  {Boolean}     [options.safe]    The safe option.
    */
-  constructor(client, name, options={}) {
+  constructor(client, name, options = {}) {
     /**
      * @ignore
      */
@@ -47,9 +46,9 @@ export default class Bucket {
    * @param  {Object} [options={}] The options to merge.
    * @return {Object}              The merged options.
    */
-  _bucketOptions(options={}) {
+  _bucketOptions(options = {}) {
     const headers = {
-      ...this.options && this.options.headers,
+      ...(this.options && this.options.headers),
       ...options.headers
     };
     return {
@@ -70,10 +69,14 @@ export default class Bucket {
    * @param  {Boolean} [options.safe]    The safe option.
    * @return {Collection}
    */
-  collection(name, options={}) {
-    return new Collection(this.client, this, name, this._bucketOptions(options));
+  collection(name, options = {}) {
+    return new Collection(
+      this.client,
+      this,
+      name,
+      this._bucketOptions(options)
+    );
   }
-
 
   /**
    * Retrieves bucket data.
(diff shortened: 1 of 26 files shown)


src/collection.js
```diff
@@ -195,16 +205,22 @@ export default class Collection {
...
-    return this.client.execute(addAttachmentRequest, {stringify: false})
+    const addAttachmentRequest = requests.addAttachmentRequest(
+      path,
+      dataURI,
+      {
+        data: record,
+        permissions
+      },
+      reqOptions
+    );
```
glasserc: The fact that this gets formatted like this is probably a good suggestion to extract the object literal.

src/collection.js
```diff
@@ -339,19 +359,20 @@ export default class Collection {
...
-      sort: "-target.data.last_modified",
-      filters: {
-        resource_name: "record",
-        collection_id: this.name,
-        "max_target.data.last_modified": String(at),
-      }
-    })
-      .then(({data: changes}) => {
+    return this.bucket
+      .listHistory({
```
glasserc: This is a strange line break..

src/collection.js
```diff
@@ -363,9 +384,11 @@ export default class Collection {
         return {
           last_modified: String(at),
           data: snapshot.sort((a, b) => b.last_modified - a.last_modified),
-          next: () => { throw new Error("Snapshots don't support pagination"); },
+          next: () => {
+            throw new Error("Snapshots don't support pagination");
+          },
           hasNextPage: false,
-          totalRecords: snapshot.length,
+          totalRecords: snapshot.length
```
glasserc: Personally I'd prefer to keep the trailing commas wherever possible.

test/batch_test.js
```diff
@@ -22,93 +21,91 @@ describe("batch module", () => {
...
-        {status: 200, body: {data: {id: 1}}},
-        {status: 201, body: {data: {id: 2}}},
+        { status: 200, body: { data: { id: 1 } } },
+        { status: 201, body: { data: { id: 2 } } }
       ];
 
-      expect(aggregate(responses, _requests))
-        .to.have.property("published")
+      expect(aggregate(responses, _requests)).to.have
+        .property("published")
```
glasserc: I'm not crazy about this, but maybe we should just get rid of `.to.have`?

n1k0:
> I'm a little nervous about this note in the Prettier documentation. Does that mean we can look forward to reformatting the project on a future prettier release?

I kinda know the folks behind, I'm not too afraid. They'll break things only if required.

> I'm not excited about what that will do to git blame.

I rarely use git blame, and when I do I usually crawl history to find who worked on a given feature at a given moment in the project history. I'm not too attached to the immediate author name attached to a given atomic commit. But yeah I understand.

I'm gonna enable trailing commas and move to 120c lines, good idea.

n1k0:
I don't like the result using 120c (see https://github.com/Kinto/kinto-http.js/pull/165/commits/7194087fc927054cde3b8e1feb1f7a5fb9a0cf94)

I'm gonna switch back to 80c.

src/bucket.js
```diff
@@ -70,23 +69,26 @@ export default class Bucket {
...
    */
-  collection(name, options={}) {
-    return new Collection(this.client, this, name, this._bucketOptions(options));
+  collection(name, options = {}) {
+    return new Collection(
+      this.client,
+      this,
+      name,
+      this._bucketOptions(options)
+    );
```
n1k0: You mean, you're not crazy about `this` in JavaScript? Yeah, me neither.

Sorry. Yeah this is slightly odd, but *at least* it's consistent :)

src/bucket.js
```diff
@@ -70,23 +69,26 @@ export default class Bucket {
...
    * @param  {Object} [options.headers] The headers object option.
    * @return {Promise<Object, Error>}
    */
-  getData(options={}) {
-    const reqOptions = {...this._bucketOptions(options)};
-    const request = {...reqOptions, path: endpoint("bucket", this.name)};
-    return this.client.execute(request)
-      .then((res) => res.data);
+  getData(options = {}) {
+    const reqOptions = { ...this._bucketOptions(options) };
```
n1k0: Sadly, or actually fortunately, prettier is opinionated and doesn't expose configuration flags for that ;)

n1k0:
I'm landing this.

PR #166, opened 2017-03-10
esdoc: remove esdoc-es7-plugin

glasserc:
Per https://github.com/esdoc/esdoc/issues/390, esdoc-es7-plugin is
deprecated and esdoc has itself grown functionality equivalent to
it. Now using esdoc-es7-plugin causes esdoc generation to fail.

Following the fix in the linked issue, just turn on all the options,
since I'm not sure which ones we use specifically.

Thanks tiberiuichim for reporting this issue on Slack.

### DIFF
--- a/esdoc.json
+++ b/esdoc.json
@@ -5,6 +5,16 @@
     "type": "mocha",
     "source": "./test"
   },
+  "experimentalProposal": {
+    "classProperties": true,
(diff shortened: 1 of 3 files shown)


glasserc:
Same as: https://github.com/Kinto/kinto.js/pull/660

PR #168, opened 2017-03-13
Install git precommit hook for linting and formatting.

n1k0:
This installs a git precommit hook for linting and formatting code against prettier configuration. 

tmp


n1k0:
I'm landing this.

PR #172, opened 2017-03-17
Fix #171: Allow creating buckets with no id specified.

### LINKED ISSUE #171: createBucket() needs to be provided with bucket id
According to the Kinto API, http://kinto.readthedocs.io/en/latest/api/1.x/buckets.html#creating-a-bucket

> Creates a new bucket. If id is not provided, it is automatically generated.

The problem is that kinto-http.js requires the bucket id parameter. It should not require it, the generated bucket id can be retrieved from the response.

n1k0:
Refs #171.


n1k0:
Released in [v4.2.0](https://github.com/Kinto/kinto-http.js/releases/tag/v4.2.0).

PR #173, opened 2017-03-20
Check for snapshot history data availability.

n1k0:
This adds checks to ensure we dispose of all the history required to compute a snapshot back to requested timestamp.

Refs https://github.com/Kinto/kinto-admin/issues/402

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -350,21 +350,7 @@ export default class Collection {
   /**
    * @private
    */
-  async findOldestRecord(at) {
-    const { data: [oldestRecord] } = await this.listRecords({
-      sort: "last_modified",
-      limit: 1,
-      filters: {
-        min_last_modified: String(at),
-      },
-    });
-    return oldestRecord;
-  }
-
-  /**
-   * @private
-   */
-  async findOldestHistoryEntry() {
+  async findHistoryOldestTimestamp() {
     const { data: [oldestHistoryEntry] } = await this.bucket.listHistory({
       sort: "target.data.last_modified",
       limit: 1,
@@ -373,7 +359,9 @@ export default class Collection {
         collection_id: this.name,
       },
     });
-    return oldestHistoryEntry;
+    return oldestHistoryEntry
+      ? oldestHistoryEntry.target.data.last_modified
+      : 0;
   }
 
   /**
@@ -399,24 +387,15 @@ export default class Collection {
     if (!Number.isInteger(at) || at <= 0) {
       throw new Error("Invalid argument, expected a positive integer.");
     }
-    // First, check that the history plugin covers the entire range back to the
-    // requested timestamp.
-    // 1. find the record timestamp in the current collection which
-    // last_modified is immediately fresher than the requested timestamp.
-    const oldestRecord = await this.findOldestRecord(at);
-    if (!oldestRecord) {
-      // If the current records list is empty, we can't find its oldest
-      // record, so we can't ensure we have enough history data to compute
-      // an accurate snapshot, which is too risky.
-      throw new Error("Cannot compute a snapshot against an empty list.");
-    }
-    // 2. find the oldest history entry for the current collection
-    const oldestHistoryEntry = await this.findOldestHistoryEntry();
-    // 3. if history is more recent, reject with an error
-    const enoughHistory = oldestHistoryEntry.target.data.last_modified <=
-      oldestRecord.last_modified;
-    if (!oldestHistoryEntry || !enoughHistory) {
-      throw new Error("Not enough history data.");
+    // If history doesn't have enough data, reject with an error
+    const oldestHistory = await this.findHistoryOldestTimestamp();
+    if (oldestHistory > at) {
+      throw new Error(
+        [
+          "Not enough history data; history for this collection",
+          `started being recorded at ${oldestHistory}.`,
+        ].join(" ")
+      );
     }
     // Retrieve history and replay it to compute the requested snapshot.
     const changes = await this.listChangesBackTo(at);
--- a/src/collection.js
+++ b/src/collection.js
@@ -361,7 +361,7 @@ export default class Collection {
     });
     return oldestHistoryEntry
       ? oldestHistoryEntry.target.data.last_modified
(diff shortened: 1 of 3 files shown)


n1k0:
This would obviously love some dedicated tests, though the good news is that current ones don't break with this patch.

I'd like feedback for this from @glasserc, @leplatrem and/or @Natim before going further. Thanks!

src/collection.js
```diff
@@ -354,17 +354,54 @@ export default class Collection {
...
-        filters: {
-          resource_name: "record",
-          collection_id: this.name,
-          "max_target.data.last_modified": String(at),
-        },
+    let oldestRecord;
+    // First, check that the history plugin covers the entire range back to the
+    // requested timestamp.
+    // 1. find the record timestamp in the current collection which
+    // last_modified is immediately fresher than the requested timestamp.
```
n1k0: I realize that what we'd need is a `created_at` field, not `last_modified`. As is, this doesn't make any sense :(

src/collection.js
```diff
@@ -354,17 +354,54 @@ export default class Collection {
...
+          },
+        });
+      })
+      .then(({ data: [oldestHistoryEntry] }) => {
+        // 3. if history is more recent, reject with an error
+        if (
+          !oldestHistoryEntry ||
+          oldestHistoryEntry.target.data.last_modified > oldestRecord
+        ) {
+          throw new Error("Not enough history data.");
```
leplatrem: We could add some attributes on the error? `error.oldestEntry` for ex.?

src/collection.js
n1k0: > What if we just skip the records that were created/modified before the oldest history entry?

I don't get it. You're at T=10, collection was created at T=0. You require a snapshot at T=2 and history started recording change events at T=5. How could you possibly infer how to reconstruct records back to T=2?

leplatrem: Yeah you're right. I was confused with diffs...

leplatrem:
I think I would only keep the check about *oldest history entry >= requested timestamp*.

src/collection.js
```diff
@@ -365,10 +389,39 @@ export default class Collection {
...
+   * @private
+   */
+  async getSnapshot(at) {
+    if (!Number.isInteger(at) || at <= 0) {
+      throw new Error("Invalid argument, expected a positive integer.");
+    }
+    // First, check that the history plugin covers the entire range back to the
+    // requested timestamp.
+    // 1. find the record timestamp in the current collection which
+    // last_modified is immediately fresher than the requested timestamp.
```
leplatrem: I don't get this part :) Why is this necessary?

src/collection.js
```diff
@@ -365,10 +389,39 @@ export default class Collection {
...
+    const oldestRecord = await this.findOldestRecord(at);
+    if (!oldestRecord) {
+      // If the current records list is empty, we can't find its oldest
+      // record, so we can't ensure we have enough history data to compute
+      // an accurate snapshot, which is too risky.
```
leplatrem: Why would it be a problem to have an empty list of records?

src/collection.js
```diff
@@ -365,10 +389,39 @@ export default class Collection {
...
+      throw new Error("Cannot compute a snapshot against an empty list.");
+    }
+    // 2. find the oldest history entry for the current collection
+    const oldestHistoryEntry = await this.findOldestHistoryEntry();
```
leplatrem: Why not use `changes[0]` here? It would get rid of `findOldestHistoryEntry()` and save a request to the server

n1k0: Because `changes[0]` is not the oldest history entry, it's the oldest history entry matching the requested range. We need to ensure history has actually been enabled *before* the requested timestamp, which information is not provided by this query.

n1k0:
Simpler approach implemented in https://github.com/Kinto/kinto-http.js/pull/173/commits/10824cfe2107650f8b9ed9338c56c59780b1145a. Feedback?

I still don't have any idea how to test for these using an actual Kinto server running...

src/collection.js
```diff
@@ -365,10 +377,32 @@ export default class Collection {
...
+   * @private
+   */
+  async getSnapshot(at) {
+    if (!Number.isInteger(at) || at <= 0) {
+      throw new Error("Invalid argument, expected a positive integer.");
+    }
+    // If history doesn't have enough data, reject with an error
+    const oldestHistory = await this.findHistoryOldestTimestamp();
+    if (!oldestHistory) {
+      throw new Error("History is not enabled.");
```
leplatrem: This is not exactly true. Empty collections may have empty records history.

Better to use a decorator that requires the history capability then!

src/collection.js
```diff
@@ -365,10 +377,32 @@ export default class Collection {
...
 
+  /**
+   * @private
+   */
+  async getSnapshot(at) {
+    if (!Number.isInteger(at) || at <= 0) {
+      throw new Error("Invalid argument, expected a positive integer.");
+    }
+    // If history doesn't have enough data, reject with an error
+    const oldestHistory = await this.findHistoryOldestTimestamp();
```
leplatrem: When I look at the 2 requests in `listChangesBackTo` and `findHistoryOldestTimestamp()` I cannot see any problem about the requested range. In one we ask for `T[0]{1}` and the other one for `T[at..0]`. In both we have the oldest entry (`T[0]`).

So here we could use `changes[changes.length -1]` (not `[0]` as I said earlier, I had missed the sort).

n1k0: In my understanding we'd rather need `changes[changes.length + 1]` which is impossible to get. Ok I'm officially burning out with this, wanna pair on it?

src/collection.js
n1k0: It was part of a future plan indeed, I just want to have a proper logic implemented first.

src/collection.js
leplatrem: We request the history from `T==0` to `T<=at` sorted descending.

src/collection.js
leplatrem: Okay, but throwing if no record history entry is not correct :)

PR #174, opened 2017-03-22
Async await

n1k0:
Migrate our current async source code to use [async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function).

Notes:
 
- The http.js module preliminary refactoring has been isolated in a [single commit](https://github.com/Kinto/kinto-http.js/pull/174/commits/23f376e8e367e82566dcdbd8a0dd1508ef373e97) to ease the review process.
- The diff is slightly larger mostly because of that large http.js module refactoring.

### DIFF
--- a/src/base.js
+++ b/src/base.js
@@ -288,7 +288,7 @@ export default class KintoClientBase {
     const maxRequests = serverSettings["batch_max_requests"];
     if (maxRequests && requests.length > maxRequests) {
       const chunks = partition(requests, maxRequests);
-      return await pMap(chunks, chunk => this._batchRequests(chunk, options));
+      return pMap(chunks, chunk => this._batchRequests(chunk, options));
     }
     const { responses } = await this.execute({
       ...reqOptions,
@@ -395,7 +395,7 @@ export default class KintoClientBase {
       if (!nextPage) {
         throw new Error("Pagination exhausted.");
       }
-      return await processNextPage(nextPage);
+      return processNextPage(nextPage);
     };
 
     const processNextPage = async function(nextPage) {
@@ -431,7 +431,7 @@ export default class KintoClientBase {
         return pageResults(results, nextPage, etag, totalRecords);
       }
       // Follow next page
-      return await processNextPage(nextPage);
+      return processNextPage(nextPage);
     };
 
     return handleResponse(
@@ -456,7 +456,7 @@ export default class KintoClientBase {
     // Ensure the default sort parameter is something that exists in permissions
     // entries, as `last_modified` doesn't; here, we pick "id".
     const paginationOptions = { sort: "id", ...options };
-    return await this.paginatedList(path, paginationOptions, reqOptions);
+    return this.paginatedList(path, paginationOptions, reqOptions);
   }
 
   /**
@@ -469,7 +469,7 @@ export default class KintoClientBase {
   async listBuckets(options = {}) {
     const path = endpoint("bucket");
     const reqOptions = this._getRequestOptions(options);
-    return await this.paginatedList(path, options, reqOptions);
+    return this.paginatedList(path, options, reqOptions);
   }
 
   /**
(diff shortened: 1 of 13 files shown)


src/utils.js
```diff
@@ -33,20 +33,16 @@ export function partition(array, n) {
...
-    )
-    .then(() => results);
+  await list.reduce(
+    async function(promise, entry) {
+      await promise;
+      results = results.concat(await fn(entry));
+    },
+    Promise.resolve()
+  );
+  return results;
```
glasserc: This might be outside the scope of this PR, but this function bothers me. It seems like it has a generic, mathy-sounding name, but we only actually use it in one very specific place (`_batchRequests`). Despite the simple name `pMap`, it's actually (to use Haskell terminology) a combined (map . bind) + sequence + concat. Despite the clean name `pMap`, there's a lot of implicit ordering and implementation details present. Maybe future work can move this function from `utils.pMap` to `base.makeChunkRequests` or something.

src/base.js
```diff
@@ -367,18 +361,17 @@ export default class KintoClientBase {
...
         "operation and should not be consumed.";
-      return Promise.resolve(raw ? { json: msg, headers: { get() {} } } : msg);
+      return raw ? { json: msg, headers: { get() {} } } : msg;
     }
-    const promise = this.fetchServerSettings().then(_ => {
-      return this.http.request(this.remote + request.path, {
-        ...request,
-        body: stringify ? JSON.stringify(request.body) : request.body,
-      });
+    await this.fetchServerSettings();
```
glasserc: This might be outside the scope of this PR, but why do we do this? Is this "priming the pump" for some other behavior that requires it?

test/bucket_test.js
```diff
@@ -507,17 +507,15 @@ describe("Bucket", () => {
...
     });
 
     it("should throw if id is missing", () => {
-      expect(() => getBlogBucket().updateGroup({})).to.Throw(
-        Error,
-        /group id is required/
-      );
+      return getBlogBucket()
+        .updateGroup({})
+        .should.be.rejectedWith(Error, /group id is required/);
```
glasserc: Nice cleanups!

src/collection.js
```diff
@@ -297,11 +297,11 @@ export default class Collection {
    * @param  {Object} [options.headers] The headers object option.
    * @return {Promise<Object, Error>}
    */
-  getRecord(id, options = {}) {
+  async getRecord(id, options = {}) {
     const path = endpoint("record", this.bucket.name, this.name, id);
     const reqOptions = this._collOptions(options);
     const request = { ...reqOptions, path };
-    return this.client.execute(request);
+    return await this.client.execute(request);
```
glasserc: It seems to me like we could have missed one of these and everything would still behave the same way. Is that correct? Is there any eslint support for making sure that any `async` function has to be `await`'ed? I guess there are use cases when you don't want to, but they seem pretty rare to me.

src/base.js
n1k0: Yes, so we're sure to fetch server settings before using them.

src/collection.js
n1k0: Thanks for wondering, that made me look into eslint docs and discover that `return await` is [always implicit](http://eslint.org/docs/rules/no-return-await), therefore unnecessary.

src/utils.js
n1k0: Yeah, we should probably do what you suggest but in a next PR.

src/base.js
glasserc: But where are we using the server settings? Not in this function, right?

n1k0: `execute` is used by every other method performing an HTTP query to the server, and I didn't want to call that function in each of them (especially as many are decorated with the `@capable` decorator, which needs server capabilities from the settings). This is basically a way to initialize the client with server settings once and for all (the `fetchServerSettings` memoizes the settings the first time it knows about them). 

How would you do it? Also, it sounds like we should comment this part of the code a little better.

PR #175, opened 2017-03-29
Improve test server crash reporting, allow using external test server

n1k0:
This patch improves Kinto test server crash reporting for better debuggability:


n1k0:
BTW this very build will fail because we've identified a regression in Kinto server combined with attachments.

PR #176, opened 2017-03-29
Feat: add and remove permission operations

gabisurita:


src/bucket.js
```diff
@@ -350,6 +350,66 @@ export default class Bucket {
...
+  async removePermissions(permissions, options = {}) {
+    if (!isObject(permissions)) {
+      throw new Error("A permissions object is required.");
+    }
+    const path = endpoint("bucket", this.name);
+    const { last_modified } = options;
+    const reqOptions = {
+      last_modified: last_modified,
+      ...this._bucketOptions(options),
+    };
```
n1k0: You can write this as 

```js
const reqOptions = { last_modified, ...this._bucketOptions(options) };
```

src/requests.js
```diff
@@ -64,6 +64,44 @@ export function updateRequest(path, { data, permissions }, options = {}) {
...
+  const {
+    headers,
+    safe,
+    last_modified,
+  } = { ...requestDefaults, ...options };
+
+  const ops = [];
+
+  for (const type in permissions) {
+    for (const principal in permissions[type]) {
```
n1k0: It's actually `const principal for permissions[type]` with arrays. It works because with `in` here we're using the indexes, but it's more semantically correct to use `for .. of`. Also you can then directly use `principal` instead of `permissions[type][principal]`.

src/requests.js
```diff
@@ -64,6 +64,44 @@ export function updateRequest(path, { data, permissions }, options = {}) {
...
+) {
+  const {
+    headers,
+    safe,
+    last_modified,
+  } = { ...requestDefaults, ...options };
+
+  const ops = [];
+
+  for (const type in permissions) {
```
n1k0: I'm wondering about adding a check for a valid permissions object here. Thoughts?

src/requests.js
```diff
@@ -64,6 +64,44 @@ export function updateRequest(path, { data, permissions }, options = {}) {
...
+    for (const principal in permissions[type]) {
+      ops.push({
+        op: opType,
+        path: "/permissions/" + type + "/" + permissions[type][principal],
```
n1k0: And so that would be 

```js
path: `/permissions/${type}/${principal}`,
```

test/bucket_test.js
```diff
@@ -759,6 +759,98 @@ describe("Bucket", () => {
...
+      sandbox.stub(requests, "jsonPatchPermissionsRequest");
+      sandbox.stub(client, "execute").returns(
+        Promise.resolve({
+          data: {},
+          permissions: fakePermissions,
+        })
+      );
+    });
+
+    it("should set permissions", () => {
```
n1k0: Nit: *should add* permissions maybe. Or even possibly *append*.

test/integration_test.js
```diff
@@ -705,43 +705,82 @@ describe("Integration tests", function() {
...
-            });
+      describe(".removePermissions()", () => {
+        beforeEach(() => {
+          return Promise.all([
+            bucket.setPermissions({ read: ["github:n1k0"] }),
+            bucket.setData({ a: 1 }),
+          ]);
+        });
+
+        it("should add bucket permissions", () => {
```
n1k0: should *remove* bucket permissions

test/integration_test.js
```diff
@@ -1347,6 +1386,47 @@ describe("Integration tests", function() {
...
+
+          describe(".removePermissions()", () => {
+            beforeEach(() => {
+              return Promise.all([
+                coll.setPermissions({ read: ["github:n1k0"] }),
+                coll.setData({ a: 1 }),
+              ]);
+            });
+
+            it("should add bucket permissions", () => {
```
n1k0: *should remove collection permissions*

test/integration_test.js
```diff
@@ -1347,6 +1386,47 @@ describe("Integration tests", function() {
...
 
+          describe(".addPermissions()", () => {
+            beforeEach(() => {
+              return Promise.all([
+                coll.setPermissions({ read: ["github:n1k0"] }),
+                coll.setData({ a: 1 }),
+              ]);
+            });
+
+            it("should add bucket permissions", () => {
```
n1k0: *should add collection permissions* (maybe *append*)

src/requests.js
```diff
@@ -64,6 +64,44 @@ export function updateRequest(path, { data, permissions }, options = {}) {
...
+      });
+    }
+  }
+
+  return {
+    method: "PATCH",
+    path,
+    headers: {
+      "Content-Type": "application/json-patch+json",
```
n1k0: I'd probably put this one at the last position so we're sure it's always applied. Here if `headers` contains a `Content-Type`, the one we need is overridden.

src/bucket.js
```diff
@@ -350,6 +350,66 @@ export default class Bucket {
...
+      path,
+      permissions,
+      "add",
+      reqOptions
+    );
+    return this.client.execute(request);
+  }
+
+  /**
+   * Remove principals to the bucket permissions.
```
n1k0: *Remove principals from the bucket permissions*

src/collection.js
```diff
@@ -170,6 +170,60 @@ export default class Collection {
...
+      path,
+      permissions,
+      "add",
+      reqOptions
+    );
+    return this.client.execute(request);
+  }
+
+  /**
+   * Remove principals to the collection permissions.
```
n1k0: *Remove principals from the collection permissions.*

src/requests.js
gabisurita: I think we could leave it to the server. Do we check for permission schemas elsewhere?

gabisurita:
Can we merge this? @n1k0

n1k0:
Yes we can :+1:  

Thank you so much and sorry for the delay!

PR #178, opened 2017-04-05
integration_test: Always initialize `server` variable

glasserc:
Otherwise, we get errors that we cannot read `flush` of undefined.

### DIFF
--- a/test/integration_test.js
+++ b/test/integration_test.js
@@ -33,9 +33,6 @@ describe("Integration tests", function() {
   this.timeout(0);
 
   before(() => {
-    if (skipLocalServer) {
-      return;
-    }
     let kintoConfigPath = __dirname + "/kinto.ini";
     if (process.env.SERVER && process.env.SERVER !== "master") {
       kintoConfigPath = `${__dirname}/kinto-${process.env.SERVER}.ini`;


test/integration_test.js
```diff
@@ -33,9 +33,6 @@ describe("Integration tests", function() {
   this.timeout(0);
 
   before(() => {
-    if (skipLocalServer) {
-      return;
-    }
```
n1k0: Well, that means always instantiating a local test server while `skipLocalServer` denotes the developer wants to use another running local instance. Ideally we should rather add a test for it before calling `server.flush()`, though that would means flushing the user local kinto server, which:

1. is a risky operation as developers don't necessarily want their kinto server instance to be emptied entirely
2. requires calling the flush endpoint manually using a http request

At that point I suspect this is getting too risky and complicated. I'd be in favor of partially reverting #175, namely https://github.com/Kinto/kinto-http.js/pull/175/commits/bd19451059a04f1f5281912fd8db79bb718a967e.

Thoughts?

glasserc: I think that's the right move, but we should expose tracebacks when they occur in the test server.

Natim:
Any status update on this one? Close? Merge?

glasserc:
An alternative approach will be pursued instead.

PR #179, opened 2017-04-05
integration_test: show how listCollections breaks

glasserc:
This appears to be the issue reported in
https://github.com/Kinto/kinto-admin/issues/421.

While we're here, add some parameters to the bucket_test to show what's going on.

I'm not sure if this issue exists in other places -- when I modified just this method in an attempt to work around the problem, I hit it in some other places. On the other hand, the tests appear to work successfully against kinto 7.0.0.

### DIFF
--- a/test/bucket_test.js
+++ b/test/bucket_test.js
@@ -176,7 +176,7 @@ describe("Bucket", () => {
         client.paginatedList,
         "/buckets/blog/collections",
         {},
-        { headers: { Foo: "Bar", Baz: "Qux" } }
+        { bucket: "blog", batch: false, headers: { Foo: "Bar", Baz: "Qux" } }
       );
     });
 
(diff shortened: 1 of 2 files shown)


n1k0:
I think we quite never considered using list operations in a batch at the time we implemented batching as it was mostly used in kinto.js to compute a result aggregate.

So I'm not sure listing operations make sense with the `aggregate` option set to `true`, we should probably add a check for that and raise an error when users attempt at doing this. Let's file a separate issue about this.

I have a feeling all other list operation methods are impacted by this bug, I don't thing create, update and delete ones are. But I may be wrong, this needs to be verified.

glasserc:
`aggregate` defaults to `false` (actually `undefined`), and is so in this test as well.

n1k0:
Yes I know, still the problem exists when people lists things and asks for an aggregate (hence why I suggest filing a separate issue about this specific, different case).

glasserc:
This got resolved as part of #180.

PR #180, opened 2017-04-07
Refactor options so we can work with Kinto 7.0.0

glasserc:
Per #178 and https://github.com/Kinto/kinto-admin/issues/421, kinto-admin broke when running against Kinto 7.0.0. Batch requests were being constructed with random arguments like `safe`, `retry`, and `batch`, which aren't request arguments but generic "options" for clients, buckets, and collections. Previously this wasn't a problem, but in Kinto 7.0.0, we now validate requests against a colander schema, and the colander schema doesn't recognize these options. Rather than hacking up a quick fix like 1. permitting the request schema in Kinto server to allow these unknown arguments or 2. explicitly `omit`ting these specific unknown arguments when generating requests, I took the opportunity to dive into the codebase and clarify options throughout the three classes, promoting everything out of `options` into member variables until there was nothing left to track using `options`, and then deleting `options`. This is a bit more verbose but is a lot more precise, which helps when reading and trying to track what options are sensible where.

This PR is designed to be read as a series of commits rather than as a giant diff.

### DIFF
--- a/src/base.js
+++ b/src/base.js
@@ -391,16 +391,16 @@ export default class KintoClientBase {
       return raw ? { json: msg, headers: { get() {} } } : msg;
     }
     await this.fetchServerSettings();
-    // FIXME: Because any fetch() option is accepted here, `request`
-    // could have lots of parameters that wouldn't be accepted as part
-    // of a batch request. This makes it easier to shoot yourself in
-    // the foot. We should probably only pass through the four
-    // parameters accepted in a batch request: method, body, path, and
-    // headers.
     const result = await this.http.request(
       this.remote + request.path,
       {
-        ...request,
+        // Limit requests to only those parts that would be allowed in
+        // a batch request -- don't pass through other fancy fetch()
+        // options like integrity, redirect, mode because they will
+        // break on a batch request.  A batch request only allows
+        // headers, method, path (above), and body.
+        method: request.method,
+        headers: request.headers,
         body: stringify ? JSON.stringify(request.body) : request.body,
       },
       {
--- a/src/base.js
+++ b/src/base.js
@@ -197,6 +197,8 @@ export default class KintoClientBase {
 
   /**
    * As _getSafe, but for "retry".
+   *
+   * @private
    */
   _getRetry(options) {
     return { retry: this._retry, ...options }.retry;
--- a/src/base.js
+++ b/src/base.js
@@ -337,9 +337,6 @@ export default class KintoClientBase {
       events: this.events,
       batch: true,
       safe: this._getSafe(options),
-      // FIXME: this doesn't actually matter, probably, since it gets
-      // passed as "default headers" in the batch?
-      headers: this._getHeaders(options),
       retry: this._getRetry(options),
     });
     let bucketBatch, collBatch;
--- a/src/base.js
+++ b/src/base.js
@@ -513,12 +513,10 @@ export default class KintoClientBase {
 
     return handleResponse(
       await this.execute(
-        // FIXME: path should override options, and we should probably
-        // not respect options.method or options.body
         // N.B.: This doesn't use _getHeaders, because all calls to
         // `paginatedList` are assumed to come from calls that already
         // have headers merged at e.g. the bucket or collection level.
-        { path: path + "?" + querystring, ...options },
+        { headers: options.headers, path: path + "?" + querystring },
         // N.B. This doesn't use _getRetry, because all calls to
         // `paginatedList` are assumed to come from calls that already
         // used `_getRetry` at e.g. the bucket or collection level.
--- a/src/base.js
+++ b/src/base.js
@@ -64,7 +64,6 @@ export default class KintoClientBase {
      */
     this.defaultReqOptions = {
       bucket: options.bucket || "default",
-      headers: options.headers || {},
       retry: options.retry || 0,
     };
 
@@ -72,6 +71,7 @@ export default class KintoClientBase {
     this._requests = [];
     this._isBatch = !!options.batch;
     this._safe = !!options.safe;
+    this._headers = options.headers || {};
 
     // public properties
     /**
@@ -178,6 +178,7 @@ export default class KintoClientBase {
     const bucketOptions = omit(this._getRequestOptions(options), "bucket");
     return new Bucket(this, name, {
       ...bucketOptions,
+      headers: this._getHeaders(options),
       safe: this._getSafe(options),
     });
   }
@@ -199,11 +200,24 @@ export default class KintoClientBase {
     return {
       ...this.defaultReqOptions,
       ...options,
-      // Note: headers should never be overriden but extended
-      headers: {
-        ...this.defaultReqOptions.headers,
-        ...options.headers,
-      },
+    };
+  }
+
+  /**
+   * Get the value of "headers" for a given request, merging the
+   * per-request headers with our own "default" headers.
+   *
+   * Note that unlike other options, headers aren't overridden, but
+   * merged instead.
+   *
+   * @private
+   * @param {Object} options The options for a request.
+   * @returns {Object}
+   */
+  _getHeaders(options) {
+    return {
+      ...this._headers,
+      ...options.headers,
     };
   }
 
@@ -233,7 +247,10 @@ export default class KintoClientBase {
     }
     const path = this.remote + endpoint("root");
     const reqOptions = this._getRequestOptions(options);
-    const { json } = await this.http.request(path, reqOptions);
+    const { json } = await this.http.request(path, {
+      ...reqOptions,
+      headers: this._getHeaders(options),
+    });
     this.serverInfo = json;
     return this.serverInfo;
   }
@@ -296,7 +313,7 @@ export default class KintoClientBase {
    */
   async _batchRequests(requests, options = {}) {
     const reqOptions = this._getRequestOptions(options);
-    const { headers } = reqOptions;
+    const headers = this._getHeaders(options);
     if (!requests.length) {
       return [];
     }
@@ -341,6 +358,9 @@ export default class KintoClientBase {
       ...this._getRequestOptions(options),
       batch: true,
       safe: this._getSafe(options),
+      // FIXME: this doesn't actually matter, probably, since it gets
+      // passed as "default headers" in the batch?
+      headers: this._getHeaders(options),
     });
     let bucketBatch, collBatch;
     if (options.bucket) {
@@ -503,6 +523,9 @@ export default class KintoClientBase {
       await this.execute(
         // FIXME: path should override options, and we should probably
         // not respect options.method or options.body
+        // N.B.: This doesn't use _getHeaders, because all calls to
+        // `paginatedList` are assumed to come from calls that already
+        // have headers merged at e.g. the bucket or collection level.
         { path: path + "?" + querystring, ...options },
         { raw: true }
       )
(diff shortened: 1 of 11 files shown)


glasserc:
This PR is still a bit WIP:

- [ ] Update documentation wherever it's no longer valid
- [ ] Update docstrings to more correctly track valid options
- [ ] Is this a major version? I changed the signature of `execute`, which I think is internal, plus `HTTP#request`, which I think is also internal.
- [ ] Possibly rethink the arguments to `paginateList`

src/bucket.js
```diff
@@ -28,39 +29,50 @@ export default class Bucket {
...
+   * @param {Object} options The options for a request.
+   * @returns {Boolean}
+   */
+  _getSafe(options) {
+    return getOptionWithDefault(options, "safe", this._safe);
+  }
+
+  /**
+   * As _getSafe, but for "retry".
+   */
```
n1k0: Please add a `@private` tag so these won't be documented as part of the public API (but maybe esdoc treats anything prefixed with _ as a private member already?)

src/bucket.js
```diff
@@ -408,10 +466,20 @@ export default class Bucket {
...
-    return this.client.batch(fn, this._bucketOptions(options));
+    let batchOptions = {
+      bucket: this.name,
+      headers: this._getHeaders(options),
+      retry: this._getRetry(options),
+      safe: this._getSafe(options),
+    };
+    if (options.aggregate) {
+      batchOptions = { ...batchOptions, aggregate: options.aggregate };
+    }
```
n1k0: Nit: We can probably directly set the `aggregate` boolean:

```js
const batchOptions = {
  bucket: this.name,
  headers: this._getHeaders(options),
  retry: this._getRetry(options),
  safe: this._getSafe(options),
  aggregate: !!options.aggregate,
};
```

src/collection.js
```diff
@@ -308,15 +357,25 @@ export default class Collection {
...
+      headers: this._getHeaders(options),
+      safe: this._getSafe(options),
+    };
+    const { last_modified } = { ...record, ...options };
+    if (last_modified) {
+      updateOptions = { ...updateOptions, last_modified };
+    }
+    if (options.patch) {
+      updateOptions = { ...updateOptions, patch: options.patch };
+    }
```
n1k0: Nit: this may just work:

```js
const { last_modified } = { ...record, ...options };
const updateOptions = {
  headers: this._getHeaders(options),
  safe: this._getSafe(options),
  last_modified,
  patch: !!options.patch,
};
```

src/collection.js
```diff
@@ -481,15 +545,21 @@ export default class Collection {
...
       bucket: this.bucket.name,
       collection: this.name,
-    });
+      headers: this._getHeaders(options),
+      retry: this._getRetry(options),
+      safe: this._getSafe(options),
+    };
+    if (options.aggregate) {
+      batchOptions = { ...batchOptions, aggregate: options.aggregate };
+    }
```
n1k0: See previous comment about directly assigning the `aggregate` property.

src/base.js
```diff
@@ -1,9 +1,9 @@
 "use strict";
 
 import {
+  getOptionWithDefault,
   partition,
   pMap,
-  omit,
```
glasserc: I already had; it's used in `requests#updateRequest`.

src/base.js
```diff
@@ -290,15 +302,22 @@ export default class KintoClientBase {
...
-      path: endpoint("batch"),
-      method: "POST",
-      body: {
-        defaults: { headers },
-        requests: requests,
+    const { responses } = await this.execute(
+      {
+        // FIXME: is this really necessary, since it's also present in
+        // the "defaults"?
+        headers: headers,
```
glasserc: There actually are (unit) tests that ensure the current behavior, which is why I left the FIXME in.

src/base.js
```diff
@@ -320,9 +340,13 @@ export default class KintoClientBase {
...
   async batch(fn, options = {}) {
     const rootBatch = new KintoClientBase(this.remote, {
-      ...this._options,
-      ...this._getRequestOptions(options),
+      events: this.events,
       batch: true,
+      safe: this._getSafe(options),
+      // FIXME: this doesn't actually matter, probably, since it gets
+      // passed as "default headers" in the batch?
+      headers: this._getHeaders(options),
```
glasserc: The only case where I think this could matter is if we ever combined headers (for example, turning `Content-Disposition: attachment` into `Content-Disposition: attachment; filename="filename.jpg"`. But our headers in Kinto never do that; we only ever add new headers, or replace existing ones. So semantically, leaving them in `defaults` should be the same as using them in all the requests, plus it leads to a smaller request. No tests break when I remove this, so I'm going to take it out.

src/base.js
```diff
@@ -364,14 +396,59 @@ export default class KintoClientBase {
...
+   * @param  {Object}  [options.headers={}]
+   *     Headers to use during all requests.
+   * @param  {Number}  [options.retry=0]
+   *     Number of times to retry each request if the server responds
+   *     with Retry-After.
+   */
   async paginatedList(path, params, options = {}) {
+    // FIXME: this is called even in batch requests, which doesn't
+    // make any sense (since all batch requests get a "dummy"
+    // response; see execute() above).
```
glasserc: I'm not sure what the right fix is, but I don't want to preclude the possibility of a more involved refactor to turn e.g. Collections into BatchCollections which have different parameters and return different things.

src/utils.js
```diff
@@ -75,6 +75,27 @@ export function omit(obj, ...keys) {
...
+ * per-object option if it's absent.
+ *
+ * @private
+ * @param {Object} options  The options passed to an arbitrary
+ *     method.
+ * @param {String} key      The option to look up.
+ * @param {Object} fallback The value to use if the key isn't present.
+ * @returns [Object]
+ */
+export function getOptionWithDefault(options, key, fallback) {
```
glasserc: I removed this function, thanks.

src/base.js
```diff
@@ -364,14 +396,59 @@ export default class KintoClientBase {
...
-    const result = await this.http.request(this.remote + request.path, {
-      ...request,
-      body: stringify ? JSON.stringify(request.body) : request.body,
-    });
+    // FIXME: Because any fetch() option is accepted here, `request`
+    // could have lots of parameters that wouldn't be accepted as part
+    // of a batch request. This makes it easier to shoot yourself in
+    // the foot. We should probably only pass through the four
+    // parameters accepted in a batch request: method, body, path, and
+    // headers.
```
glasserc: The natural fix here is to select only those properties from `request` using something like

```
    const result = await this.http.request(
      this.remote + request.path,
      {
        method: request.method,
        headers: request.headers,
        body: stringify ? JSON.stringify(request.body) : request.body,
      },
      {
        retry: this._getRetry(options),
      }
    );
```

But this means that if `request` doesn't provide those things, we'll get `undefined` instead of missing fields. That seems a little ugly to me. What do you think?

test/api_test.js
```diff
@@ -154,9 +149,11 @@ describe("KintoClient", () => {
...
       };
 
-      expect(api.bucket("foo", options)).to.have
-        .property("options")
-        .eql(options);
+      const bucket = api.bucket("foo", options);
+      expect(bucket).property("_safe", options.safe);
+      expect(bucket).property("_retry", options.retry);
+      expect(bucket).property("_headers").eql(options.headers);
+      expect(bucket).property("_isBatch", options.batch);
```
glasserc: I don't think it would be useful to expose them publicly. The tests are trying to make sure that the fields are "recognized"; this seems like a simple way to do that, even if it involves poking around in the internals of the class. (Otherwise I don't see any way to write the test at all.)

src/bucket.js
glasserc: There are some tests using `calledWith` which (therefore) do a deep-equals equality on the resulting method arguments. See for example https://github.com/Kinto/kinto-http.js/blob/master/test/collection_test.js#L572. Unconditionally passing `aggregate` broke those tests. I didn't want to change them unnecessarily so I did this instead, although I agree it's a bit ugly. Do you think I should update the tests instead?

src/collection.js
glasserc: Same thing as above with `calledWith`. But I don't mind updating the tests if you think that's better.

src/base.js
```diff
@@ -511,9 +596,15 @@ export default class KintoClientBase {
       throw new Error("A bucket id is required.");
     }
     const path = endpoint("bucket", bucketObj.id);
-    const { last_modified } = { bucketObj };
-    const reqOptions = this._getRequestOptions({ last_modified, ...options });
-    return this.execute(requests.deleteRequest(path, reqOptions));
+    const { last_modified } = { ...bucketObj, ...options };
+    return this.execute(
+      requests.deleteRequest(path, {
```
leplatrem: nit: for consistency with other calls, this `path` might have to go next line? or others go on the same line :)

src/bucket.js
n1k0: Yes we can definitely update the tests here.

test/api_test.js
n1k0: Yeah, I was just wondering. Let's keep them private.

src/base.js
n1k0: Indeed, though let's keep that work for a (possible) next PR.

src/base.js
n1k0: Several options here:

```js
// 1. the imperative way

const {method, headers, body} = request;
const finalRequest = {};
if (method) {
  finalRequest.method = method;
}

// 2. the costly serialization way

const {method, headers, body} = request;
// stringify will drop undefined properties
const finalRequest = JSON.parse(JSON.stringify({method, headers, body}));

// 3. the dedicated utility way
const {method, headers, body} = request;
function cleanUndefinedProperties(obj) {
  const result = {};
  for (const key in obj) {
    if (typeof obj[key] !== undefined) {
      result[key] = obj[key];
    }
  }
  return result;
}
const finalRequest = cleanUndefinedProperties(request);
```

None are great, so I don't know. I think I could live with undefined request properties passed to `request` anyway, as I think `fetch` implementations all deal just fine with this situation.

src/base.js
```diff
@@ -216,8 +214,15 @@ export default class KintoClientBase {
...
-    const reqOptions = this._getRequestOptions(options);
-    const { json } = await this.http.request(path, reqOptions);
+    const { json } = await this.http.request(
+      path,
+      {
+        headers: this._getHeaders(options),
+      },
+      {
+        retry: this._getRetry(options),
+      }
```
glasserc: If `npm run cs-format` says they are, then they are...

src/base.js
```diff
@@ -290,15 +294,22 @@ export default class KintoClientBase {
...
+    const { responses } = await this.execute(
+      {
+        // FIXME: is this really necessary, since it's also present in
+        // the "defaults"?
+        headers: headers,
+        path: endpoint("batch"),
+        method: "POST",
+        body: {
+          defaults: { headers },
+          requests: requests,
```
glasserc: Oops, thanks!

src/base.js
glasserc: I just do what `npm run cs-format` says...

glasserc:
I think I'll defer any changes to the `paginatedList` interface for now in the interests of getting this out the door. I think right now it's still a patch version, since the only method APIs I changed (`HTTP#request` and `KintoClientBase#execute`) are internal. Final review and/or thoughts?

src/base.js
n1k0: Note that you can slightly influence the formatter, eg. if you write the object on a single line it will preserve that style. And yeah I wish it were fully idempotent.

src/base.js
n1k0: > I think fetch implementations all deal just fine with this situation.

Well, I was wrong. #183

PR #183, opened 2017-04-11
Fix undefined request options passed to fetch.

n1k0:
This fixes an issue preventing request to be handled properly in Google Chrome. Chrome expects fetch options like `method` to be always defined. We may pass `undefined` instead, which results in this:

 

I know @glasserc loves JavaScript, so I'd really like him to review this patch.

### DIFF
--- a/src/base.js
+++ b/src/base.js
@@ -1,6 +1,14 @@
 "use strict";
 
-import { partition, pMap, qsify, support, nobatch, toDataBody } from "./utils";
+import {
+  partition,
+  pMap,
+  qsify,
+  support,
+  nobatch,
+  toDataBody,
+  cleanUndefinedProperties,
+} from "./utils";
 import HTTP from "./http";
 import endpoint from "./endpoint";
 import * as requests from "./requests";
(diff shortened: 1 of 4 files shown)


n1k0:
I'll handle review to @Natim as I forgot @glasserc was away this aft.

src/base.js
```diff
@@ -396,7 +404,7 @@ export default class KintoClientBase {
         method: request.method,
         headers: request.headers,
         body: stringify ? JSON.stringify(request.body) : request.body,
-      },
+      }),
```
glasserc: Sorry about that! I probably would have chosen something more like this:

```
{
    ...selectProperties(request, ["method", "headers"]),
    body: stringify ? JSON.stringify(request.body) : request.body,
}
```

with

```
function selectProperties(obj, props) {
    let ret = {};
    for (const prop of props) {
        if (obj.hasOwnProperty(prop)) {
            ret[prop] = obj[prop];
        }
    }
    return ret;
}
```

But I guess this is fine too.

PR #184, opened 2017-04-12
Fix missing babel-polyfill.

n1k0:
Fix https://github.com/Kinto/kinto/issues/1190

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,7 @@
 "use strict";
 
+import "babel-polyfill";
+
 import { EventEmitter } from "events";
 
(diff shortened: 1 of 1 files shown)


leplatrem:
I tried to upgrade the kinto-admin and got this:

screenshot from 2017-04-12 13-23-26

n1k0:
It looks like babel-poyfill should be a dependency of the local admin
project as well. Could you try that?

Le 12 avr. 2017 1:24 PM, "Mathieu Leplatre" <notifications@github.com> a
écrit :

Natim:
Yes I will do that.

PR #186, opened 2017-04-17
Cleanup serverInfo a little bit

### LINKED ISSUE #181: `fetchServerInfo` + headers + memoization
`fetchServerInfo` stores a member variable `serverInfo` which is the value of the first "server info" response it got. This saves having to rerequest server info. However, `fetchServerInfo` accepts a `headers` option. Does `headers` change the behavior of the server? If so, we shouldn't rely on memoization in this way. If not, we shouldn't accept a `headers` option.

glasserc:
Try to clean up some of the warts around the use of `serverInfo`. In particular:

- Don't memoize the result of `fetchUser`, since it can change in the presence of authentication information.
- Forbid the passing of `headers` to `fetchServerInfo`, since it doesn't make any sense (the server information shouldn't change according to the presence of authentication information).
- Get rid of a needless `fetchServerSettings` in `execute()`.
- Pass `retry` to one call to `fetchServerSettings` that didn't have it.

This doesn't completely clean it up; `retry` isn't passed in calls made via `capable()` or `support()`, and I don't see a way to make it be. Honestly I'd prefer to get rid of `retry` as an option on every single method, but that's a separate discussion.

Fixes #181 and responds to https://github.com/Kinto/kinto-http.js/pull/174#discussion_r107751608.

### DIFF
--- a/src/base.js
+++ b/src/base.js
@@ -415,7 +415,8 @@ export default class KintoClientBase {
       this._requests.push(request);
       // Resolve with a message in case people attempt at consuming the result
       // from within a batch operation.
-      const msg = "This result is generated from within a batch " +
+      const msg =
+        "This result is generated from within a batch " +
         "operation and should not be consumed.";
       return raw ? { json: msg, headers: { get() {} } } : msg;
     }
--- a/src/base.js
+++ b/src/base.js
@@ -417,7 +417,6 @@ export default class KintoClientBase {
         "operation and should not be consumed.";
       return raw ? { json: msg, headers: { get() {} } } : msg;
     }
-    await this.fetchServerSettings();
     const result = await this.http.request(
       this.remote + request.path,
       cleanUndefinedProperties({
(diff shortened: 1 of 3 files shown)


n1k0:
> Get rid of a needless fetchServerSettings in execute().

The purpose of first fetching the `/hello` endpoint before executing any other request to the Kinto server was to ensure retrieving server settings like `batch_max_requests` (which is required to create a proper batch request).

Obviously this was only required for `batch()`, but meanwhoile we added the `readonly` setting, which we may want to handle as well.

Thoughts?

glasserc:
For `batch_max_requests`, we already do that (and I didn't change it) here in `_batchRequests`: https://github.com/Kinto/kinto-http.js/blob/master/src/base.js#L302

glasserc:
For `readonly`, I think we should handle this the same way as we do for capabilities/`capable` -- we should add a decorator (or what are they called in JavaScript?) that can call `fetchServerSettings` itself to discover the `readOnly` property. See https://github.com/Kinto/kinto-http.js/blob/master/src/utils.js#L185.

Issue #188, opened 2017-04-27
setData doesn't support patch anymore.


Natim:
Refs https://github.com/Kinto/kinto-admin/issues/430

Natim:
Released in kinto-http.js@4.3.3

PR #189, opened 2017-04-27
Make sure to handle patch option for collection setData.

Natim:
Refs #188

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ Read the [API documentation](https://doc.esdoc.org/github.com/Kinto/kinto-http.j
   - [Contributing](#contributing)
      - [Coding style](#coding-style)
      - [Integration tests](#integration-tests)
+     - [Releasing](#releasing)
 
 ---
 
@@ -1594,3 +1595,8 @@ It's possible to run the integration test suite against an external Kinto server
 ```
 $ TEST_KINTO_SERVER=https://my.kinto-server.tld/v1 npm test
 ```
+
+### Releasing
+
+If you need to release a new version of kinto-http.js, you can follow
+[the Release documentation](https://github.com/Kinto/kinto-http.js/wiki/Releasing).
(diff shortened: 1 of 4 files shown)


test/collection_test.js
```diff
@@ -278,12 +278,32 @@ describe("Collection", () => {
...
+        requests.updateRequest,
+        "/buckets/blog/collections/posts",
+        {
+          data: { a: 1 },
+          permissions: undefined,
+        },
+        {
+          headers: { Foo: "Bar", Baz: "Qux" },
+          patch: true,
+          safe: false,
```
Natim: I am not sure about safe: false and last_modiifed: undefined here.

Natim:
Tests fails but Travis is happy. https://travis-ci.org/Kinto/kinto-http.js/jobs/226332789

src/collection.js
```diff
@@ -156,6 +156,7 @@ export default class Collection {
       { data, permissions },
       {
         last_modified,
+        patch: !!patch,
```
leplatrem: I know that I'm the one who suggested you to put `patch: false` when unspecified, but I see that is not the way it's done elsewhere :/ eg. https://github.com/Kinto/kinto-http.js/blob/master/src/bucket.js#L150

README.md
```diff
@@ -1594,3 +1595,8 @@ It's possible to run the integration test suite against an external Kinto server
 ```
 $ TEST_KINTO_SERVER=https://my.kinto-server.tld/v1 npm test
 ```
+
+### Releasing
+
+If you need to release a new version of kinto-http.js, you can follow
+[the Release documentation](https://github.com/Kinto/kinto-http.js/wiki/Releasing).
```
glasserc: 👍

leplatrem:
> This was probably my fault, sorry!

I wouldn't say so... if there would have been tests for `collection.setData()` you would have spotted it ;)

PR #195, opened 2017-05-30
Remove form-data dependency

glasserc:
This appears not to cause tests to fail.


n1k0:
Sadly we get a lot of `HTTP 400; Error: HTTP 400 Bad Request: Invalid request parameter (Content-Type should be multipart/form-data)` failure. I think that's why I used this specific package instead of relying on JSDom's impl.

glasserc:
That's really strange, I could have sworn it ran locally without errors. OK, thanks!

PR #255, opened 2017-11-14
Require babel-polyfill only when not already loaded.

francois2metz:


Natim:
Thank you for that. I will let @glasserc sign-off but it looks good :)

glasserc:
Refs #159. I'm not convinced we have the correct behavior here, but I don't think this makes anything any worse..

PR #260, opened 2017-11-17
Use transform-runtime to request regenerator without depending on it

### LINKED ISSUE #259: Kinto is not defined with version 4.4.0
cc @Natim @glasserc 

Hey, I had to lock to 4.3.4 in the Notes PR: https://github.com/mozilla/notes/pull/313/commits/f9257c89fc07afa90a759a736324d2ad41da49b3

Otherwise it just throw `Kinto is not defined` with 4.4.0

image

glasserc:
[transform-runtime](https://babeljs.io/docs/plugins/transform-runtime) is meant to allow us to access the regenerator
polyfill "seamlessly, without global pollution". This means we don't
need to import it explicitly, and we don't need to make any efforts to
import it only once.

This replaces babel-polyfill, which is meant for applications, not
libraries.

Applications are responsible for providing the polyfill.

We use the "polyfill: false" option to disable conversion of "builtin"
JS methods such as Array.from and Object.create, which we could also
rewrite except for
https://github.com/loganfsmyth/babel-plugin-transform-builtin-extend/issues/15.

This is meant to fix #259 and Kinto/kinto#1379.

Refs #159 (cc: @screendriver).

Refs Kinto/kinto-admin#456 and Kinto/kinto-admin#446.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,7 +8,10 @@
           "globals": ["Error"]
         }],
         "transform-object-rest-spread",
-        "transform-async-to-generator"
+        "transform-async-to-generator",
+        ["transform-runtime", {
+          "polyfill": false
+        }]
       ]
     },
     "firefox": {
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
   },
   "homepage": "https://github.com/Kinto/kinto-http.js#readme",
   "dependencies": {
+    "babel-runtime": "^6.22.0",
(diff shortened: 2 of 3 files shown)


Natim:
Shall we try to build the kinto-admin with it?

glasserc:
How do you mean? I've already tried it locally..

Natim:
> I've already tried it locally..

r+ then :)

Issue #225, opened 2017-08-09
Remove transform-object-rest-spread from Gecko build


leplatrem:
Rest/spread is now supported in FFx55 https://bugzilla.mozilla.org/show_bug.cgi?id=1339395

leplatrem:
```
> kinto-http@4.4.1 dist-fx /home/mathieu/Code/Mozilla/kinto-client
> BABEL_ENV=firefox browserify -s KintoHttpClient --ignore events --bare -e fx-src/index.js -o temp.jsm -t [ babelify --sourceMapRelative . ] && mkdir -p dist && cp fx-src/jsm_prefix.js dist/moz-kinto-http-client.js && echo "
/*
 * Version $npm_package_version - $(git rev-parse --short HEAD)
 */
" >> dist/moz-kinto-http-client.js && cat temp.jsm >> dist/moz-kinto-http-client.js && rm temp.jsm

SyntaxError: /home/mathieu/Code/Mozilla/kinto-client/fx-src/index.js: Unexpected token (34:28)
  32 |     const events = {};
  33 |     EventEmitter.decorate(events);
> 34 |     super(remote, { events, ...options });
     |                             ^
  35 |   }
  36 | }
  37 | 
    at Parser.pp$5.raise (/home/mathieu/Code/Mozilla/kinto-client/node_modules/babylon/lib/index.js:4454:13)
    at Parser.pp.unexpected (/home/mathieu/Code/Mozilla/kinto-client/node_modules/babylon/lib/index.js:1761:8)
    at Parser.pp$3.parseIdentifier (/home/mathieu/Code/Mozilla/kinto-client/node_modules/babylon/lib/index.js:4332:10)
    at Parser.pp$3.parsePropertyName (/home/mathieu/Code/Mozilla/kinto-client/node_modules/babylon/lib/index.js:4156:96)
    at Parser.pp$3.parseObj (/home/mathieu/Code/Mozilla/kinto-client/node_modules/babylon/lib/index.js:4045:12)
    at Parser.pp$3.parseExprAtom (/home/mathieu/Code/Mozilla/kinto-client/node_modules/babylon/lib/index.js:3719:19)
    at Parser.pp$3.parseExprSubscripts (/home/mathieu/Code/Mozilla/kinto-client/node_modules/babylon/lib/index.js:3494:19)
[...]

PR #273, opened 2018-02-12
Firefox updates

glasserc:
Another attempt to build a version of this library for inclusion in Firefox, plus taking care of #225 now that all the dependencies are in place. To make this transition smoother, we make some changes here to mirror changes that have been made to our compiled output in the Firefox source tree.

### DIFF
--- a/fx-src/index.js
+++ b/fx-src/index.js
@@ -18,11 +18,9 @@
 import KintoClientBase from "../src/base";
 import * as errors from "../src/errors";
 
-const Cu = Components.utils;
-
-Cu.import("resource://gre/modules/Timer.jsm");
+ChromeUtils.import("resource://gre/modules/Timer.jsm");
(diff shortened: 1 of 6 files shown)


fx-src/index.js
```diff
@@ -18,11 +18,9 @@
 import KintoClientBase from "../src/base";
 import * as errors from "../src/errors";
 
-const Cu = Components.utils;
-
-Cu.import("resource://gre/modules/Timer.jsm");
+ChromeUtils.import("resource://gre/modules/Timer.jsm");
```
leplatrem: I didn't know :) 

https://bugzilla.mozilla.org/show_bug.cgi?id=1425611

glasserc: Yeah, it's a bit of a surprise, but an announcement was posted to dev-platform/firefox-dev.

PR #278, opened 2018-04-03
Fix testing

glasserc:
Combines with https://github.com/Kinto/kinto/pull/1576 to fix the current test failures in Travis.

Also, while we're here, fix the test warnings from the sinon version.

### DIFF
--- a/test/api_test.js
+++ b/test/api_test.js
@@ -21,7 +21,7 @@ describe("KintoClient", () => {
   let sandbox, api, events;
 
   beforeEach(() => {
-    sandbox = sinon.sandbox.create();
+    sandbox = sinon.createSandbox();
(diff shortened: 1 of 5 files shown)


glasserc:
@leplatrem Because we now expose `bucket:create` in master, but not in 6.0.1, the same set of tests cannot pass in both configurations. What do you think we should do? Strip out that one and use the old assertions? Enforce that it's there but only on `master`?

leplatrem:
The only thing that comes into my mind (it's late here) is to have a condition on the server version in the tests.
`if serverInfo.version < 8.3 then skip bucket:create`

glasserc:
OK, ready for a final review on this. @leplatrem ?

PR #289, opened 2018-06-07
Replace then() with await in README

leplatrem:


README.md
```diff
@@ -198,8 +198,7 @@ kinto.setHeaders({
 A Kinto server exposes some of its internal settings, information about authenticated user, the HTTP API version and the API capabilities (e.g. plugins).
 
 ```js
-client.fetchServerInfo([options])
-  .then(({data}) => ...);
+const info = await client.fetchServerInfo([options]);
```
magopian: Should it also be `const {data} = await ...` like below?

README.md
```diff
@@ -246,8 +245,7 @@ Sample result:
 ### Listing buckets
 
 ```js
-client.listBuckets([options])
-  .then(({data}) => ...);
+const { data } = await client.listBuckets([options]);
```
magopian: You wrote it `{data}` (without the spaces) through the rest of the file, not sure what Prettier would want.

README.md
leplatrem: well, no, it was a mistake, there's no `data` attribute

PR #304, opened 2018-10-02
Tests use accounts

glasserc:
Since Kinto/kinto#1736 landed, the test suite for both this library and kinto.js have been failing constantly (see e.g. #300, #301, #302, #303). Here is a fix which moves all tests to using accounts which should fix it.

Note that with this fix, the test suite takes a lot longer -- over 20 minutes on my laptop. I'm not exactly sure why but I bet it's due to the increased need for cryptographically hashing passwords once per request. Of special note is `should handle long list of changes`, which took almost 2 minutes on my machine. Maybe @leplatrem can make it faster on master, and maybe we should leave it on basicauth for the older kintos?

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ const kinto = new KintoClient("https://my.server.tld/v1", {
 
 > #### Notes
 >
-> - As explained in the [server docs](http://kinto.readthedocs.io/en/stable/api/1.x/authentication.html#basic-auth), any string is accepted. You're not obliged to use the `username:password` format.
+> - As Kinto Accounts has become the default as of 11, you may need to create an account first.
 
 ### Using an OAuth Bearer Token
 
--- a/test/integration_test.js
+++ b/test/integration_test.js
@@ -87,7 +87,9 @@ describe("Integration tests", function() {
       return stopServer(server);
     });
 
-    beforeEach(() => server.flush());
+    beforeEach(() => {
+      return server.flush().then(_ => api.createAccount("user", "pass"));
+    });
 
     // XXX move this to batch tests
     describe("new batch", () => {
@@ -152,7 +154,7 @@ describe("Integration tests", function() {
 
       it("should retrieve user information", () => {
         return api.fetchUser().then(user => {
-          expect(user.id).to.match(/^basicauth:/);
+          expect(user.id).to.equal("account:user");
           expect(user.bucket).to.have.length.of(36);
         });
       });
@@ -319,6 +321,8 @@ describe("Integration tests", function() {
               expect(bucketCreate.length).eql(1);
               data = data.filter(p => !isBucketCreate(p));
             }
+            const isOwnAccount = p => p.uri == "/accounts/user";
+            data = data.filter(p => !isOwnAccount(p));
             expect(data).to.have.length.of(2);
             expect(data.map(p => p.id).sort()).eql(["b1", "c1"]);
           });
@@ -340,6 +344,7 @@ describe("Integration tests", function() {
             if (shouldHaveCreatePermission) {
               expectedRecords++;
             }
+            expectedRecords++; // for account:create
             expect(results.data).to.have.length.of(expectedRecords);
           });
         });
@@ -575,7 +580,9 @@ describe("Integration tests", function() {
 
     after(() => stopServer(server));
 
-    beforeEach(() => server.flush());
+    beforeEach(() => {
+      return server.flush().then(_ => api.createAccount("user", "pass"));
+    });
 
     it("should appropriately populate the backoff property", () => {
       // Issuing a first api call to retrieve backoff information
@@ -586,8 +593,6 @@ describe("Integration tests", function() {
   });
 
   describe("Deprecated protocol version", () => {
-    beforeEach(() => server.flush());
-
     describe("Soft EOL", () => {
       before(() => {
         const tomorrow = new Date(new Date().getTime() + 86400000)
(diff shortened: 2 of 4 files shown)


src/base.js
```diff
@@ -684,4 +684,15 @@ export default class KintoClientBase {
       { retry: this._getRetry(options) }
     );
   }
+
+  @capable(["accounts"])
+  async createAccount(username, password) {
```
leplatrem: We should mention this in the README

README.md
```diff
@@ -163,7 +163,7 @@ const kinto = new KintoClient("https://my.server.tld/v1", {
 
 > #### Notes
 >
-> - As explained in the [server docs](http://kinto.readthedocs.io/en/stable/api/1.x/authentication.html#basic-auth), any string is accepted. You're not obliged to use the `username:password` format.
+> - As Kinto Accounts has become the default as of 10.2.0, you may need to create an account first.
```
leplatrem: Probably v11

src/requests.js
```diff
@@ -30,8 +30,9 @@ export function createRequest(path, { data, permissions }, options = {}) {
     ...requestDefaults,
     ...options,
   };
+  const method = options.method || (data && data.id) ? "PUT" : "POST";
   return {
-    method: data && data.id ? "PUT" : "POST",
+    method: method,
```
leplatrem: method is superflous

test/kinto-8.1.2.ini
```diff
@@ -24,6 +20,11 @@ kinto.attachment.base_path = /tmp
 # Enable permissions endpoint
 kinto.experimental_permissions_endpoint = True
 
+# Force accounts (not default until 10.2)
+multiauth.policies = account
+multiauth.policy.account.use = kinto.plugins.accounts.authentication.AccountsAuthenticationPolicy
```
leplatrem: `kinto.plugins.accounts.AccountsPolicy`

.travis.yml
```diff
@@ -5,7 +5,7 @@ env:
 - ACTION="run lint"
 - ACTION="run dist" # ensures building dist files doesn't break
 - ACTION="run cs-check"
-- SERVER=6.0.1 ACTION=test
+- SERVER=8.1.2 ACTION=test
```
leplatrem: some old version on purpose?

glasserc:
Refs also Kinto/kinto#1821.

.travis.yml
glasserc: I don't understand. 8.1.2 *is* the old version on purpose. 7.0.0 is the oldest one that supports `accounts`, but I didn't see a reason to keep anything from that era around.

glasserc:
Rebased onto current master and cleaned up a little bit since #312 landed. This PR now serves to move tests to accounts and stop using basicauth in tests. This will still make running the test suite much more expensive, but fixing Kinto/kinto#1821 will have the same effect so maybe we should just go for it to ensure that we're testing a scenario that matches the Kinto "default".

README.md
```diff
@@ -165,7 +165,7 @@ const kinto = new KintoClient("https://my.server.tld/v1", {
 
 > #### Notes
 >
-> - As explained in the [server docs](http://kinto.readthedocs.io/en/stable/api/1.x/authentication.html#basic-auth), any string is accepted. You're not obliged to use the `username:password` format.
+> - As Kinto Accounts has become the default as of 11, you may need to create an account first.
```
leplatrem: *as of version 11* ?

test/integration_test.js
```diff
@@ -87,7 +87,9 @@ describe("Integration tests", function() {
       return stopServer(server);
     });
 
-    beforeEach(() => server.flush());
+    beforeEach(() => {
+      return server.flush().then(_ => api.createAccount("user", "pass"));
```
itaisteinherz:
```suggestion
      return server.flush().then(() => api.createAccount("user", "pass"));
```

test/integration_test.js
```diff
@@ -319,6 +321,8 @@ describe("Integration tests", function() {
               expect(bucketCreate.length).eql(1);
               data = data.filter(p => !isBucketCreate(p));
             }
+            const isOwnAccount = p => p.uri == "/accounts/user";
+            data = data.filter(p => !isOwnAccount(p));
```
itaisteinherz: Would be simpler to remove the line above and change this to:

```suggestion
            data = data.filter(p => p.uri !== "/accounts/user");
```

Also, `p` is not a very clear variable name, and so it would be preferable to use a longer and more readble one.

test/integration_test.js
```diff
@@ -575,7 +580,9 @@ describe("Integration tests", function() {
 
     after(() => stopServer(server));
 
-    beforeEach(() => server.flush());
+    beforeEach(() => {
+      return server.flush().then(_ => api.createAccount("user", "pass"));
```
itaisteinherz:
```suggestion
      return server.flush().then(() => api.createAccount("user", "pass"));
```

(Same as https://github.com/Kinto/kinto-http.js/pull/304/files#r249295696)

test/integration_test.js
```diff
@@ -646,7 +651,9 @@ describe("Integration tests", function() {
...
```
itaisteinherz:
```suggestion
      return server.flush().then(() => api.createAccount("user", "pass"));
```

(Same as https://github.com/Kinto/kinto-http.js/pull/304/files#r249295696)

test/integration_test.js
```diff
@@ -680,7 +687,9 @@ describe("Integration tests", function() {
...
```
itaisteinherz:
```suggestion
      return server.flush().then(() => api.createAccount("user", "pass"));
```

(Same as https://github.com/Kinto/kinto-http.js/pull/304/files#r249295696)

leplatrem:
Closing. Reopen if willing to see this landing :)

PR #308, opened 2018-10-08
Add support for query arguments to Collection#getData

### LINKED ISSUE #307: Additional querystring parameter for expected timestamp
In the context of Firefox Remote Settings, we run Kinto behind a CDN and we would like to do cache busting via querystring parameters.

For example, when we fetch the collection metadata, we want the URL to be something like this: `/v1/buckets/bid/collections/cid/records?_expected=12346789`

I was about to suggest a new generic `queryParams` option for `.getData()` but @glasserc suggested we could also introduce the notion of expected timestamp (and maybe raise if the one obtained differs or something).

> Note 1: For cache busting in listing records, we can already use the `filters` option `{filters: {_expected: "42"}}
> Note 2: this would apply to every resource, not only collection metadata.

glasserc:
Ref #307.

This represents a minimal change that fixes the issue. In particular, it doesn't add support for `_fields` (which should probably be supported as a separate `option`) and it doesn't add support to query arguments for buckets, records, or groups (are there other resources too?). The goal of this PR is to land just this feature for the purposes of getting uplifted.

### DIFF
--- a/src/collection.js
+++ b/src/collection.js
@@ -3,6 +3,7 @@ import { v4 as uuid } from "uuid";
 import { capable, toDataBody, isObject } from "./utils";
 import * as requests from "./requests";
 import endpoint from "./endpoint";
+import { qsify } from "./utils";
 
 /**
  * Abstract representation of a selected collection.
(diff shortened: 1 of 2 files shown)


glasserc:
N.B. I wasn't really sure what to call the option. It isn't really a "filter" since it only applies to one record..

PR #312, opened 2018-10-17
Add accounts support

glasserc:
This is meant to extract the part of #304 that adds a `createAccount` method, and nothing besides that.

I had a horrible realization when I was writing this that when adding the `accounts` plugin was going to make the entire test suite slow, even for those tests that do `basicauth`. Fortunately, it doesn't; see https://github.com/Kinto/kinto/issues/1821. Still, we might want to hold off on merging this until we figure out our plan for that bug.

Another option might be to use a separate config for just the account tests. But then we might have to have two, one for each Kinto version..

### DIFF
--- a/src/requests.js
+++ b/src/requests.js
@@ -30,8 +30,9 @@ export function createRequest(path, { data, permissions }, options = {}) {
     ...requestDefaults,
     ...options,
   };
+  const method = options.method || (data && data.id) ? "PUT" : "POST";
   return {
-    method: data && data.id ? "PUT" : "POST",
+    method: method,
     path,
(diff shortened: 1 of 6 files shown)


src/requests.js
```diff
@@ -30,8 +30,9 @@ export function createRequest(path, { data, permissions }, options = {}) {
     ...requestDefaults,
     ...options,
   };
+  const method = options.method || (data && data.id) ? "PUT" : "POST";
   return {
-    method: data && data.id ? "PUT" : "POST",
+    method: method,
```
leplatrem: can be written `method,`

PR #313, opened 2018-10-17
Support more queries

### LINKED ISSUE #307: Additional querystring parameter for expected timestamp

glasserc:
Fixes #307.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ Read the [API documentation](https://doc.esdoc.org/github.com/Kinto/kinto-http.j
      - [Sorting](#sorting)
      - [Polling for changes](#polling-for-changes)
      - [Paginating results](#paginating-results)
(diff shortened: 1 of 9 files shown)


glasserc:
Oh, yeah, I should update the README. Sorry, I looked for a CHANGELOG and when I didn't see one, I just sort of gave up.

PR #317, opened 2018-11-29
Restore use of versions.txt when installing master

glasserc:
This got dropped soon after it got added, but it seems like this is
why we fetch versions.txt. It's my hope that putting this back in will
cause tests to pass (right now they're failing because kinto relies on
jsonschema==3.0.0a3 and we're installing 2.6.0).


Natim:
Ok nice catch!

PR #351, opened 2019-01-29
 Support for retrieving ETag (#128)

### LINKED ISSUE #128: Better way for client programs to do a HEAD request to a collection endpoint
If they want to get the record count in a collection, for instance.

Here is current way of doing it (getting the api from a kinto.js collection):

```
var api = tickets.api.bucket(tickets.bucket).collection(tickets.name);
var headers = api._collOptions({});
headers["Authorization"] = "Basic " + btoa("token:tr0ub4d@ur");
api.client.execute({ path: "/buckets/default/collections/oldTickets/records", method:"HEAD", headers}, {raw:true})
```

kewisch:
* Add a getETag method for collections and buckets

    * Update README to document methods and no longer suggest use of last_modified

This fixes #128 . I'm not quite sure how ETag and last modified work here, the server we are using says I cannot pass the last_modified. The docs say it should be treated as an opaque token, which doesn't seem to be true. The linked issue says it is hard to get the ETag, so I'd like to make it simpler. Comments welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Read the [API documentation](https://doc.esdoc.org/github.com/Kinto/kinto-http.j
      - [Updating an existing group](#updating-an-existing-group)
      - [Deleting a group](#deleting-a-group)
      - [Listing bucket history](#listing-bucket-history)
+     - [Bucket ETag](#bucket-etag)
   - [Collections](#collections)
      - [Selecting a collection](#selecting-a-collection)
      - [Getting collection data](#getting-collection-data)
@@ -50,6 +51,7 @@ Read the [API documentation](https://doc.esdoc.org/github.com/Kinto/kinto-http.j
      - [Deleting record](#deleting-record)
      - [Listing records](#listing-records)
      - [Total number of records](#total-number-of-records)
+     - [Collection ETag](#collection-etag)
      - [Batching operations](#batching-operations)
   - [Listing all resource permissions](#listing-all-resource-permissions)
   - [Attachments](#attachments)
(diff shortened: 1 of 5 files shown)


leplatrem:
Thanks for starting this! 
*disclaimer* I haven't read the code yet ;)


>  the server we are using says I cannot pass the last_modified

Could you elaborate on that please? Do you mean via this JS client?

I think we should call it *timestamp* (`getTimestamp()`?), since we call it like that on the server API docs. What do you think?

kewisch:
What I actually wanted to do is get entries that are in blocklists-preview but not blocklists. So I found this in the docs:
>- `last_modified`: the [collection's timestamp](http://kinto.readthedocs.io/en/stable/api/1.x/timestamps.html). This value is opaque and should be reused as is, eg. passing it as a `since` option (see [Generic bucket and collection options](#generic-bucket-and-collection-options))

So I tried:

```js
let data = await this.bucket("blocklist").collection("addons").getData();
let preview = await this.bucket("blocklist-preview").collection("addons").listRecords({ since: data.last_modified });
```

And the server replied with: `Invalid value for since (1548760721714), should be ETag value`

So I need some way to get the ETag instead. If I instead pass the ETag from a HEAD request it works fine. Am I missing something?

leplatrem:
Thanks for the details! :pray: 

So there is indeed some confusion in our docs... 

In brief, there are two timestamps in play here: 
- the timestamp of the collection object (bumped when the collection metadata is modified)
- the timestamp of the records list (bumped when a record is created/modified/deleted in that collection)

Both are bumped independently, and are not interchangeable. 


Basically, what you want to achieve could be done like this:
```js
const { last_modified } = await this.bucket("blocklists").collection("addons").listRecords();  // inefficient GET just for a header value... 
const preview = await this.bucket("blocklists-preview").collection("addons").listRecords({ since: last_modified });
```


In the python client ([kinto-http.py](https://github.com/Kinto/kinto-http.py/)), we only added one method called `get_records_timestamp()`. I believe we should mimic this in the JS client, so that your snippet becomes:

```js
const last_modified = await this.bucket("blocklists").collection("addons").getRecordsTimestamp(); // HEAD
const preview = await this.bucket("blocklists-preview").collection("addons").listRecords({ since: last_modified });
```

leplatrem:
*Oops, didn't want to publish so fast...*

Does it make sense?

If we would like to generalize the idea to every kind of resources, we would have methods like `getGroupsTimestamp()`, `getCollectionsTimestamp()` in the bucket object.

Thank you for your patience ;)

leplatrem:
Superseded by #425

PR #425, opened 2019-06-01
Retrieve timestamps (ETag) of plural endpoints

leplatrem:
Follow-up of #351

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,15 @@ Read the [API documentation](https://doc.esdoc.org/github.com/Kinto/kinto-http.j
      - [Deleting a bucket](#deleting-a-bucket)
      - [Creating a collection](#creating-a-collection)
      - [Listing bucket collections](#listing-bucket-collections)
+     - [Collections list timestamp](#collections-list-timestamp)
      - [Deleting a collection](#deleting-a-collection)
      - [Creating a user group](#creating-a-user-group)
      - [Listing bucket groups](#listing-bucket-groups)
+     - [Groups list timestamp](#groups-list-timestamp)
      - [Getting a bucket group](#getting-a-bucket-group)
      - [Updating an existing group](#updating-an-existing-group)
      - [Deleting a group](#deleting-a-group)
      - [Listing bucket history](#listing-bucket-history)
-     - [Bucket ETag](#bucket-etag)
   - [Collections](#collections)
      - [Selecting a collection](#selecting-a-collection)
      - [Getting collection data](#getting-collection-data)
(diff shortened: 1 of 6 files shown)


leplatrem:
Tests fail because of #424

README.md
```diff
@@ -1168,10 +1191,31 @@ Sample result:
 42
 ```
 
-#### Options
+#### options
```
glasserc: I think this should remain capitalized?

README.md
```diff
@@ -1142,7 +1165,7 @@ Sample result:
 
 The result object exposes the following properties:
 
-- `last_modified`: the [collection's timestamp](http://kinto.readthedocs.io/en/stable/api/1.x/timestamps.html). This value is opaque and should be reused as is, eg. passing it as a `since` option (see [Generic bucket and collection options](#generic-bucket-and-collection-options))
+- `last_modified`: the [collection's timestamp](http://kinto.readthedocs.io/en/stable/api/1.x/timestamps.html).
```
glasserc: Maybe we should note that it is not the same as the timestamps you get from `getCollectionsTimestamp()` etc.

README.md
```diff
@@ -1168,10 +1191,31 @@ Sample result:
 42
 ```
 
-#### Options
+#### options
 
-- `headers`: Custom headers object to send along the HTTP request
-- `retry`: Number of retries when request fails (default: 0)
+- `headers`: custom headers object to send along the http request
```
glasserc: I think HTTP could remain capitalized.

README.md
```diff
@@ -552,6 +555,26 @@ Sample result:
 
 This method accepts the [generic parameters for sorting, filtering and paginating results](#generic-options-for-list-operations).
 
+### Collections list timestamp
+
+The timestmap of the collections list is used for the `since` option in the [generic parameters for sorting, filtering and paginating results](#generic-options-for-list-operations).
```
glasserc: You have `timestamp` typoed as `timestmap` a few places.

src/bucket.js
```diff
@@ -93,6 +93,52 @@ export default class Bucket {
...
+    };
+    const { headers } = await this.client.execute(request, {
+      raw: true,
+      retry: this._getRetry(options),
+    });
+    return headers.get("ETag");
+  }
+
+  /**
+   * Retrieves the ETag of the collection list, for use with the `since` filtering option.
```
glasserc: `of the group list`?

README.md
leplatrem: Actually, it is the same :) I'll add a note

PR #445, opened 2019-07-05
Gecko: move import of global objects to JSM prefix

leplatrem:


fx-src/jsm_prefix.js
```diff
@@ -21,3 +21,7 @@
 const global = this;
 
 var EXPORTED_SYMBOLS = ["KintoHttpClient"];
+
+const { setTimeout, clearTimeout } = ChromeUtils.import("resource://gre/modules/Timer.jsm");
+const { XPCOMUtils } = ChromeUtils.import("resource://gre/modules/XPCOMUtils.jsm");
+XPCOMUtils.defineLazyGlobalGetters(global, ["fetch"]);
```
glasserc: Missing newline

Natim:
Can you explain why?

leplatrem:
> Can you explain why? 

Oh :) Yes, it's a matter of closure. 

Before we were doing `ChromeUtils.import("resource://gre/modules/Timer.jsm", global);`, so it could be done anywhere/anytime.

But now, since we do `const { setTimeout } = ChromeUtils.import("resource://gre/modules/Timer.jsm");`, it has to be global, ie. for the whole file

Issue #468, opened 2019-07-23
TypeError: Blob is not a constructor


ncloudioj:
Ran into this error when using kinto-http.js to upload attachments as follows:

```js
async function publish(collection) {
  const fileStream = fs.createReadStream(`${L10N_REPO_PATH}/all-locales`);
  const rl = readline.createInterface({
    input: fileStream,
    crlfDelay: Infinity,
  });

  for await (const line of rl) {
    const ftlPath = `${L10N_REPO_PATH}/${line}/browser/browser/newtab/asrouter.ftl`;
    if (fs.existsSync(ftlPath)) {
      const dataUri = await DataURI(ftlPath);
      try {
        await collection.addAttachment(dataUri, {id: `cfr-${VERSION}-${line}`}, {filename: "asrouter.ftl"});
      }
      catch (e) {
        console.error(e);
      }
    } else {
      console.log("Missing Fluent file for ", line);
    }
  }
}
```

The stacktrace:

```
TypeError: Blob is not a constructor
    at extractFileInfo (/Users/.../node_modules/kinto-http/lib/utils.js:459:14)
    at createFormData (/Users/.../node_modules/kinto-http/lib/utils.js:483:26)
    at Object.addAttachmentRequest (/Users/.../node_modules/kinto-http/lib/requests.js:209:44)
    at Collection._callee9$ (/Users/.../node_modules/kinto-http/lib/collection.js:664:49)
    at tryCatch (/Users/.../node_modules/regenerator-runtime/runtime.js:45:40)

```

glasserc:
Hmm. Looks like kinto-http.js expects to be run in a browser environment. The tests run in NodeJS using `Buffer.from` (https://github.com/Kinto/kinto-http.js/blob/master/test/setup-jsdom.js#L18). I wonder if we can support both environments using something like this. It seems like the only place we use this functionality is in adding `attachment` to a `formData`. Let me try to write a PR.

glasserc:
Closing in favor of #470, "Should we shim other browser dependencies".

PR #469, opened 2019-07-23
If Blob is not present, use Buffer

glasserc:
Attempt to address #468. There may be other NodeJS problems, but let's take them one at a time.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -279,7 +279,14 @@ export function extractFileInfo(dataURL) {
   for (let i = 0; i < binary.length; i++) {
     array.push(binary.charCodeAt(i));
   }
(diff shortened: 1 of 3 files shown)


ncloudioj:
Yes, it works now (with the two extra shims for "atob" and "FormData")!

Thank y'all!

glasserc:
I guess we can merge this since it makes things better and it's a pretty easy fix. @ncloudioj suggests that we should shim `atob`, `FormData`, and `fetch` (using `node-fetch`). If this doesn't get merged, I'll try to add those to this PR as I figure out how.

leplatrem:
I don't know if it is the responsibility of a library to shim everything from browser.

leplatrem:
Let's merge this and discuss about additional shims elsewhere!

PR #472, opened 2019-07-29
Fix #471: obfuscate authorization header values in errors

### LINKED ISSUE #471: Don't include authorization information in exceptions
@rfk points out in https://bugzilla.mozilla.org/show_bug.cgi?id=1566529 that log messages including credentials is probably bad.

leplatrem:
Fix #471

There were many ways to solve this... I could have obfuscated the value in the error constructor, could have removed the options from the logging message, etc. 
Let me know what do you think!

### DIFF
--- a/test/http_test.js
+++ b/test/http_test.js
@@ -130,13 +130,30 @@ describe("HTTP class", () => {
     });
 
     describe("Request timeout", () => {
-      it("should timeout the request", () => {
+      beforeEach(() => {
         sandbox.stub(global, "fetch").returns(
           new Promise(resolve => {
             setTimeout(resolve, 20000);
           })
         );
+      });
+
+      it("should timeout the request", () => {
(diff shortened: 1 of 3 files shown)


test/http_test.js
```diff
@@ -130,14 +130,31 @@ describe("HTTP class", () => {
...
+        return http
+          .request("/", {
+            mode: "cors",
+            headers: {
+              Authorization: "XXX",
+              "User-agent": "mocha-test",
+            },
+          })
+          .should.be.rejectedWith(
+            'Timeout while trying to access / with {"mode":"cors","headers":{"Accept":"application/json","Content-Type":"application/json","Authorization":"XXX","User-agent":"mocha-test"}}'
```
glasserc: We have this test that shows that the Authorization header is shown, but we don't have a test showing that the Authorization header is actually redacted?

PR #476, opened 2019-08-08
Add initial set of TypeScript types

dstaley:
This PR adds TypeScript types to the following files:

1. `batch.ts`
1. `endpoint.ts`
1. `errors.ts`
1. `utils.ts`

In the process of adding types, I made a few changes to better support static type-checking:

1. The `pMap` function was replaced with an inline version that relies on the `async`/`await` syntax along with a for loop. I personally think this version is more readable, but it's also more able to be type-checked. `pMap` was only used in one location in `base.js`.
1. The endpoint functions were changed from `endpoint("routename")` to `endpoint.routename()`. This allowed for type-checking the route parameters.
1. The `replaceKey` utility function was replaced with a more specific `obscureAuthorizationHeader` function that works regardless of the format the user specifies headers. Headers for `fetch` can take one of three forms: a multi-dimensional array, an object, and a native `Headers` object. `obscureAuthorizationHeader` works with all three of these formats.
1. The `omit` utility function was replaced with a faster inline implementation. `omit` was only used in one location in `requests.js`.

Also, `isomorphic-fetch` was replaced with `node-fetch` since the latter more closely matches what's available in evergreen browsers today. (`isomorphic-fetch` hasn't seen an update in four years.)


src/requests.js
```diff
@@ -46,7 +46,10 @@ export function updateRequest(path, { data, permissions }, options = {}) {
   const { headers, safe, patch } = { ...requestDefaults, ...options };
   const { last_modified } = { ...data, ...options };
 
-  if (Object.keys(omit(data, "id", "last_modified")).length === 0) {
+  if (
+    Object.keys(data).filter(k => k !== "id" && k !== "last_modified")
```
leplatrem: I would use an intermediary variable here, for better readibility

PR #478, opened 2019-08-11
Add types to http and requests

dstaley:
These types are fairly straight-forward. The only snag I ran into was that some request types might rely on the `data` parameter having a `last_modified` property. I'm fairly sure these are all invoked with a record since they're all operations on an existing record, but if that's not the case I'll revise these types in the next round of changes.


src/requests.ts
```diff
@@ -1,5 +1,23 @@
...
+  method?: HttpMethod;
+  gzipped?: boolean | null;
+  last_modified?: number;
+}
+type AddAttachmentRequestOptions = RequestOptions & {
+  last_modified?: number;
+  filename?: string;
+};
+
+type RequestBody = { data?: any; permissions?: Record<string, string[]> };
```
leplatrem: I guess that `Record` is the `Map` type for typescript?

dstaley: Yes! Although after thinking about it a bit more I realized I wasn't using `Record` correctly here. I pushed an update that defines `permissions` as `Partial<Record<Permission, string[]>>`, along with a new `Permission` type that defines the allowed keys based on the API documentation. This allows for type-checking permission objects, ensuring that you don't pass an invalid permission. As I work through the rest of the library, I'll be able to refine the type in circumstances where it makes sense (for example, only allowing `read` and `write` on a record).

PR #479, opened 2019-08-13
Convert tests to TypeScript

dstaley:
This PR converts the tests for the existing TypeScript files to TypeScript. Writing the tests in TypeScript gives us the ability to test functionality and the type system. It allows us to ensure that a given change doesn't break compilation on existing TypeScript code.

Moving forward, I'll convert the tests to TypeScript in the same PR as the main file conversion.


src/endpoint.ts
```diff
@@ -12,7 +12,7 @@ const ENDPOINTS = {
     `${ENDPOINTS.bucket(bucket)}/collections` + (coll ? `/${coll}` : ""),
   group: (bucket: string, group?: string) =>
     `${ENDPOINTS.bucket(bucket)}/groups` + (group ? `/${group}` : ""),
-  record: (bucket: string, coll: string, id?: string) =>
+  record: (bucket: string, coll: string, id?: string | number) =>
```
leplatrem: I wouldn't allow number here

src/types.ts
```diff
@@ -9,7 +9,7 @@ export interface KintoRequest {
 
 export interface KintoRecord {
   id: string;
-  last_modified: number;
+  last_modified?: number;
```
leplatrem: If the record comes from the server, it will always have this field. Which part of the tests required this?

dstaley: Particularly when updating a record, we know the ID, but we don't pass in a `last_modified` inside the record. I've added a new type, `KintoIdRecord`, which is used when we know a record's ID, but don't want to specify a `last_modified`. `KintoRecord` now extends from `KintoIdRecord`, adding the `last_modified` property.

leplatrem:
Awesome! Thanks!

PR #480, opened 2019-08-15
Add types to KintoClientBase

dstaley:
Alright, this is where the PRs start getting a little tricky.

This PR adds types to `KintoClientBase`. The majority of the added types are straight-forward duplications of the types already existing in the JSDoc comments. However, there are a few places I'd like to call out for some additional attention:

1. A generic type parameter was added to the request-related functions in `http.ts`. This allows the caller to specify the datatype of the returned data, allowing for additional type safety.
1. `KintoClientBase.execute` has two response types: `HttpResponse<T>` and `T`. These map to the `raw` parameter being `true` and `false` respectively. I'd really like to remove the `raw` parameter completely, thus making `execute` always return `HttpResponse<T>`. This would remove the need to cast the return down to `T` since the return type would always be `HttpResponse<T>`. Since `bucket.js` and `collection.js` rely on the `raw` parameter, I didn't make that change in this PR, but once those are converted to TypeScript I'd like to make that change.
1. `partition` was updated to always return `T[][]` as opposed to returning `T[]` when the length was less than the maximum. This means that `KintoClientBase. _batchRequests` can always be called with `KintoRequest[]`. This was the only use of `partition`.
1. The typings for `KintoClientBase.batch` will likely be refined once I finish converting `bucket.js` and `collection.js`. Right now it accepts a function that accepts one of `KintoClientBase`, `Bucket`, or `Collection`, but it might be a good candidate for an interface type, that way the caller doesn't have to check which class the argument is an instance of.
1. Some new server response types were added to `types.ts`. These were based off the documentation and the return values for my Kinto server. They compile, meaning they cover all the properties accessed in `kinto-http`, but if you notice some missing please let me know!
1. There were a few instances where the JSDoc comments didn't match how certain parameters were being used. In these cases, I defaulted to making the necessary changes to compile. For example, the JSDoc for `KintoClientBase._batchRequests` didn't specify a `retry` parameter, but the function called `this._getRetry` passing in the provided options.


src/base.ts
```diff
@@ -66,7 +107,17 @@ export default class KintoClientBase {
...
+    try {
+      version = remote.match(/\/(v\d+)\/?$/)![1];
+    } catch (err) {
+      throw new Error("The remote URL must contain the version: " + remote);
+    }
+    if (version !== SUPPORTED_PROTOCOL_VERSION) {
+      throw new Error(`Unsupported protocol version: ${version}`);
+    }
+    this._remote = remote;
+    this._version = version;
```
leplatrem: Why did you have to inline this code instead of using the setter?

src/base.ts
```diff
@@ -624,7 +752,16 @@ export default class KintoClientBase {
    *     when faced with transient errors.
    * @return {Promise<Object, Error>}
    */
-  async createBucket(id, options = {}) {
+  async createBucket(
+    id: string | null,
+    options: {
+      data?: any;
```
leplatrem: Shouldn't it be `Object` here? (ie. it can't be a string or a number...)

src/http.ts
```diff
@@ -124,11 +125,11 @@ export default class HTTP {
   /**
    * @private
    */
-  async processResponse(response: Response): Promise<HttpResponse> {
+  async processResponse<T>(response: Response): Promise<HttpResponse<T>> {
     const { status, headers } = response;
     const text = await response.text();
     // Check if we have a body; if so parse it as JSON.
-    let json;
+    let json: unknown;
```
leplatrem: I guess it should be `T` here

src/types.ts
```diff
@@ -23,3 +23,46 @@ export type Permission =
...
+
+interface User {
+  id: string;
+  principals: string[];
+  bucket: string;
+}
+
+interface ServerCapability {
+  description: string;
+  url: string;
```
leplatrem:
```suggestion
  url: string;
  version: string;
```

src/base.ts
```diff
@@ -656,7 +793,15 @@ export default class KintoClientBase {
    * @param  {Number}        [options.last_modified] The last_modified option.
    * @return {Promise<Object, Error>}
    */
-  async deleteBucket(bucket, options = {}) {
+  async deleteBucket(
+    bucket: string | KintoIdRecord,
```
leplatrem: If this «id record» type has to be used in operations on buckets, then we should call it `KintoIdObject`

src/http.ts
dstaley: Well at this point in the code, it's not known whether `json` is `T` or `ServerResponseObject`. By leaving it `unknown`, we can prevent any unsafe access until we determine which type it might be. If we switch it to `T`, we'll need to cast it back to `unknown` before casting it to `ServerResponseObject`. I think it makes a little more sense to leave it unknown, but I'll defer to your judgement there!

src/types.ts
dstaley: Based on the response from `https://kinto.dev.mozaws.net/v1/`, not all capabilities have the `version` property. What we could do, though, is define it as optional, that way if someone does a check like `if (capability.version)` the TypeScript compiler will know it's a `string` without the user needing to cast it to one.

src/base.ts
dstaley: I introduced a new type, `MappableObject` that will reject if someone passes in an array. Unfortunately TypeScript doesn't have a built-in "object, but not array" type.

src/base.ts
dstaley: I have a feature called "strict initialization checks" enabled that requires class properties to be assigned in the constructor. However, we were actually always assigning anyway, it was just that the compiler couldn't determine that. I learned about a feature called the definite assignment assertion, so I removed the inline function and used those instead to tell the compiler that we are, in all circumstances, assigning those properties.

src/types.ts
Natim: Yes I would say that version is optional, it is part of the API. The only one that don't have a version property are plugins that ships with kinto itself.

dstaley: Ah, gotcha! In that case I think it makes sense to have it defined. Since it's an optional field, users will need to check that it's defined before accessing it, but they won't need to check that it's a string.

leplatrem:
:clap: :clap:

PR #487, opened 2019-08-24
Add types to Bucket

dstaley:
This one is fairly straightforward. It's primarily adding types to the `options` parameter for all the various methods and utilizing sinon's stubs to appease the TypeScript compiler.

There's also a few minor type adjustments:
1. `permissions` is now defined as `{ [key in Permission]?: string[] }` instead of `Record<string, string>` in `jsonPatchPermissionsRequest`. This allows us to be more strict in what values we allow into functions that pass permissions.
1. `endpoint.collection` was requiring a `collection` parameter. It is now correctly an optional parameter.
1. The parameters for `paginatedList` were pulled into their own interface so it can be shared with the other functions that pass options to it.

One thing I have a question about: is the pagination parameter `_since` or `since`? Searching through `kinto.js` I see [references](https://github.com/Kinto/kinto.js/blob/7102a48e2adb4bbcf6513fcdb6e9abbba0c0a0c6/test/collection_test.js#L2074) to `since`, but the [API documentation](https://docs.kinto-storage.org/en/stable/api/1.x/collections.html?highlight=_since#list-bucket-collections) uses `_since`. It looks like the parameter name should be `since`, but the actual query param should be `_since` (this is also corroborated by [this line](https://github.com/Kinto/kinto-http.js/blob/master/src/base.ts#L606)). I've adjusted the tests to pass `since`, but if this is incorrect please let me know.


leplatrem:
The query parameter is  `_since`. But indeed, the parameter name in the signatures is `since`

PR #492, opened 2019-08-27
Enable down-level iteration

dstaley:
`Headers.entries()` returns an `IterableIterator<[string, string]>`. Iterators are an ES6 feature, so the TypeScript compiler has to be explicitly told to convert them to something ES5 environments can understand. This PR enables that option, `downlevelIteration` in `tsconfig`.

Also, to prevent something like this from happening again, `npm run build` has been added to the list of actions for Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ python:
  - "3.6"
 env:
 - ACTION="run lint"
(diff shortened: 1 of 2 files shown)


leplatrem:
Thanks!

PR #494, opened 2019-08-28
Add types to Collection and index

dstaley:
This PR is pretty much just adding types to the methods on `Collection`.

The only notable change is the renaming of `BucketResponse` to `KintoEntity`, representing an object with a `data` property that's a `KintoRecord` (which itself is just an object with `id` and `last_modified`) and `permissions`, used when accessing the permissions of a `Collection` or `Bucket`, and also in `HistoryEntry` which is the return type for the history endpoint.

This also includes a few housekeeping changes. Now that the entire source is TypeScript, eslint had to be reconfigured a smidge. I removed the unused locals eslint rule and replaced that with the equivalent compiler option in TypeScript

I also took the opportunity to convert `index.js` and `index.fx.js` to TypeScript, meaning that once this PR is merged the entire codebase (excluding the tests) will be written in TypeScript! The remaining test files are actually the largest files in the project, so it'll be a few days before I finish converting those.


leplatrem:
You're amazing! Thanks!

PR #503, opened 2019-09-03
Convert API and integration tests to TypeScript

dstaley:
This chonker of a PR converts the API and integration tests to TypeScript. The primary changes to the library were adding spots for generic type arguments so that we can tell the compiler what types we're expecting.

Some highlights:
1. The code for invoking the batch function was tweaked a smidge to be a bit clearer. This allows us to have a strongly typed batch client (whereas before we'd need a typecast each time you invoke a method on the client argument).
1. I added a type declaration file for `kinto-node-test-server`. Once this and the main Kinto libraries are written in TypeScript, I might take a look at converting that one too, but for now a declaration file is fine especially since it's only used for testing.
1. You'll see a lot of casts (`as KintoEntity`) in the integration tests. This is due to the fact that `KintoClientBase.execute` is polymorphic. I have an issue open [here](Kinto/kinto.js#1754) tracking all the functions that have different return types based on their inputs. Once this is merged, changing `execute` to always return `HttpResponse<T>` is my first priority. So the casts are in there now, but I'll be able to remove them in the next PR.
1. I might need to take a step back and think of better names for some of the types. We now have `KintoIdObject`, `KintoRecord`, and `KintoEntity`. These represent objects with IDs, objects with IDs and a `last_modified`, and an object with a `data` property along with `permissions`. I'm open to suggestions, especially a replacement for `KintoRecord` that makes it clearer that it's for buckets, collections, and records.
1. We should also review all the public/private properties of each class. I defaulted to private, and marked them as public once they were used in tests, but it would be good to review whether we want specific properties to be public. This will be especially important once TypeScript implements support for private class fields, since private fields will be marked as such even in a browser.
1. For some reason, `.length.of()` wasn't working for me. I checked the chai docs, and it seems like it should be `.lengthOf`. I'm not sure if `length.of` was deprecated or what, but TypeScript refused to compile it. I swapped it out with `.lengthOf` and all's good.


src/base.ts
```diff
@@ -74,13 +82,13 @@ interface PaginationResult<T> {
...
   private _requests: KintoRequest[];
-  private _isBatch: boolean;
+  public _isBatch: boolean;
   private _retry: number;
-  private _safe: boolean;
-  private _headers: Record<string, string>;
-  private serverInfo: HelloResponse | null;
-  private events: EventEmitter;
-  private http: HTTP;
+  public _safe: boolean;
```
leplatrem: could be private

src/base.ts
```diff
@@ -74,13 +82,13 @@ interface PaginationResult<T> {
...
+  public _headers: Record<string, string>;
```
leplatrem: may require setter/getter if tests rely on it :/

src/base.ts
```diff
@@ -74,13 +82,13 @@ interface PaginationResult<T> {
 export default class KintoClientBase {
   private _backoffReleaseTime: number | null;
   private _requests: KintoRequest[];
-  private _isBatch: boolean;
+  public _isBatch: boolean;
```
leplatrem: could be private

leplatrem (approved):
Sorry about the super long delay :( 

Thanks for the detailed description, great as usual!

> I might need to take a step back and think of better names for some of the types. We now have KintoIdObject, KintoRecord, and KintoEntity

`KintoIdObject` sounds fine.
`KintoRecord` would be `KintoObject`
`KintoEntity` could be `KintoResponse` since this is usually what's returned by the server.

Does it make sense?

> We should also review all the public/private properties of each class. 

IMO, we should follow the convention `_*` are private, others are public

dstaley:
I also took the opportunity to mark a few class methods that both began with `_` and were labeled as `@private` as actual private functions.

leplatrem (approved):
:+1: 

It's a pity to have getters just for unit tests, but it's not a big deal...

PR #520, opened 2019-09-28
Update build scripts to support Windows

dstaley:
This PR modifies the scripts in `package.json` to use `cross-env` and `shx` to support cross-platform building (primarily adding support for Windows). `cross-env` is used to set environment variables, and `shx` is used to make commands like `cp` work on Windows.

There's two interesting changes here:
1. The `dist:fx` script has been moved out into its own script file in `bin/dist-fx.ts`. It was complicated enough that it made sense to move it out to something that was more readable. Thanks to `shelljs`, this script is essentially a 1:1 copy of the existing build script, but now it's written in TypeScript as opposed to straight shell.
2. I had to change one integration test so that it passed on Windows. On [this line](https://github.com/Kinto/kinto-http.js/pull/520/files#diff-3dc21efa3a3651d7d32bcd4dcb1ae6f8L882), we're asking Kinto to sort history objects by the `date` property, which is assigned in Python using `datetime.now().isoformat()`. I discovered that on Windows, this function only has millisecond accuracy, meaning that any operations performed in less than a millisecond end up having the exact same timestamp, which means that the sort order isn't guaranteed to be what the client expected. Since `last_modified` is more like a revision count, I switched to using that instead.

Other minor changes include specifying LF line endings for prettier, and updating the `cs-check` and `cs-format` scripts to validate TypeScript files (whoops!).


package.json
```diff
@@ -9,23 +9,24 @@
     "build": "npm run build:es && npm run build:cjs-es5",
     "build:es": "tsc -p .",
     "build:cjs-es5": "tsc -p . --outDir lib/cjs-es5 --module commonjs --target es5",
-    "build:demo": "npm run dist && cp dist/kinto-http.min.js demo/kinto-http.min.js && cp dist/kinto-http.min.js.map demo/kinto-http.min.js.map",
+    "build:demo": "npm run dist && shx cp dist/kinto-http.min.js demo/kinto-http.min.js && shx cp dist/kinto-http.min.js.map demo/kinto-http.min.js.map",
     "build:readme": "./node_modules/.bin/toctoc -w -d 2 README.md",
-    "cs-check": "prettier -l '{src,test}/**/*.js'",
-    "cs-format": "prettier '{src,test}/**/*.js' --write",
+    "cs-check": "prettier -l \"{src,test,bin}/**/*.{js,ts}\"",
```
Natim: What did motivate this change? Do you find `\"` more readable than `'` or is it because Windows shell doesn't handle `'`? :fearful:

dstaley:
The default windows shell doesn't handle single quotes unfortunately.
________________________________

package.json
dstaley: The Windows command prompt doesn't treat single quote characters the same way as double quotes unfortunately.

dstaley:
@leplatrem now that the fixes for `kinto-node-test-server` have been published this is ready for a review!

Issue #95, opened 2016-05-26
Delete on plural endpoints


leplatrem:
The server supports `DELETE` on plural endpoints (`/buckets`, `/collections`, `/records`).

It supports the same filtering options as `listRecords()`.

It can be useful to delete multiple entries with one request:
- `DELETE /buckets` : delete every buckets for which I have write permission
- `DELETE /buckets?in_id=test,prueba,essai` : delete buckets with specified ids
- `DELETE /buckets/blog/collections?since="42"`: delete every collections in bucket whose timestamp > 42 for which I have write permission
- `DELETE /buckets/blog/collections/articles/records`: delete every records...

lavish205:
This is pretty interesting, I would love to work on this.

paullaffitte:
I would like to work on this too, since @lavish205 asked for it two years ago, I guess he is not working on it, so it will not be a duplicate work.

Natim:
Sure go for it.

leplatrem:
Fixed in #534

PR #534, opened 2019-10-14
Delete on plural endpoints

paullaffitte:
It implements deleteRecords, deleteCollections and deleteBuckets, with the same filtering as listRecords. #95

### DIFF
--- a/src/base.ts
+++ b/src/base.ts
@@ -754,7 +754,7 @@ export default class KintoClientBase {
     params: PaginatedParams = {},
     options: { headers?: Record<string, string>; retry?: number } = {}
   ): Promise<PaginationResult<T>> {
-    return this.paginatedOperation(path, params, options);
+    return this.paginatedOperation<T>(path, params, options);
   }
 
   /**
@@ -805,7 +805,7 @@ export default class KintoClientBase {
       safe: safe ? safe : false,
       last_modified,
     });
-    return this.paginatedOperation(path, params, {
+    return this.paginatedOperation<T>(path, params, {
       ...options,
       headers: deleteRequest.headers as Record<string, string>,
       method: "DELETE",
--- a/src/base.ts
+++ b/src/base.ts
@@ -28,6 +28,7 @@ import {
   ServerCapability,
   User,
   Emitter,
+  HttpMethod,
 } from "./types";
 import Collection from "./collection";
 
@@ -48,7 +49,7 @@ export interface KintoClientOptions {
   batch?: boolean;
 }
 
-export interface PaginatedListParams {
+export interface PaginatedParams {
   sort?: string;
   filters?: Record<string, string | number>;
   limit?: number;
@@ -574,10 +575,11 @@ export default class KintoClientBase {
   }
 
   /**
-   * Fetch some pages from a paginated list, following the `next-page`
-   * header automatically until we have fetched the requested number
-   * of pages. Return a response with a `.next()` method that can be
-   * called to fetch more results.
+   * Perform an operation with a given HTTP method on some pages from
+   * a paginated list, following the `next-page` header automatically
+   * until we have processed the requested number of pages. Return a
+   * response with a `.next()` method that can be called to perform
+   * the requested HTTP method on more results.
    *
    * @private
    * @param  {String}  path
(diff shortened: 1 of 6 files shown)


src/collection.ts
```diff
@@ -596,6 +596,60 @@ export default class Collection {
...
+    options: PaginatedParams & {
+      headers?: Record<string, string>;
+      retry?: number;
+      // at?: number;
+    } = {}
+  ): Promise<PaginationResult<T>> {
+    const path = endpoint.record(this.bucket.name, this.name);
+    // if (options.at) {
+    //   return this.getSnapshot<T>(options.at);
+    // } else {
```
leplatrem: leftovers?

src/collection.ts
```diff
@@ -596,6 +596,60 @@ export default class Collection {
...
+    return this.client.paginatedDelete<T>(path, options, {
+      headers: this._getHeaders(options),
+      retry: this._getRetry(options),
+    });
+    // }
```
leplatrem: same ;)

src/bucket.ts
```diff
@@ -398,6 +398,33 @@ export default class Bucket {
...
+      fields?: string[];
+    } = {}
+  ): Promise<PaginationResult<KintoObject>> {
+    const path = endpoint.collection(this.name);
+    return this.client.paginatedDelete<KintoObject>(path, options, {
+      headers: this._getHeaders(options),
+      retry: this._getRetry(options),
+    });
+  }
+
```
leplatrem: We should probably add `deleteGroups()` now that we're at this :)

leplatrem:
@paullaffitte do you need any help? Let us know :pray: !

paullaffitte:
@leplatrem Sorry I was on vacation, I think it's okay, maybe I'll need help for the tests but I'll let you know. I'll try to finish it this week-end.

paullaffitte:
I'm done with writing tests, but I can't seems to find how to properly run the integration tests.
`npm run test` gives me: "Uncaught Error: spawn pserve ENOENT"
and `TEST_KINTO_SERVER=localhost:8888/v1 npm run test`: "TypeError: Cannot read property 'flush' of undefined"

dstaley:
@paullaffitte You'll need to be in a python venv to run the integration tests. (Alternatively you can `pip install kinto kinto-attachment`, but using a venv is highly recommended)

The following should work with Python 3 on Linux:

```sh
$ python3 -m venv .venv
$ source .venv/bin/activate
(.venv) $ pip install kinto kinto-attachment
$ npm run test
```

paullaffitte:
That's very nice that there is some automated tests now. If they were there at the time I started the PR, it would be already merge maybe haha.

Anyway, sorry about the time I took to complete it, I was overwhelmed by work/studies/personal projects and I kind of forgot about this PR. I rebased my branch today and updated what should be updated. I also have tested it locally but I guess that the automated tests will show us if my work is good or not.

src/base.ts
```diff
@@ -605,10 +607,14 @@ export default class KintoClientBase {
    *     Number of times to retry each request if the server responds
```
leplatrem, line 607: Add missing `method` no? With default value (cf. call from `paginateList`)

src/base.ts
```diff
@@ -708,6 +715,101 @@ export default class KintoClientBase {
...
+  async paginatedList<T>(
+    path: string,
+    params: PaginatedParams = {},
+    options: { headers?: Record<string, string>; retry?: number } = {}
+  ): Promise<PaginationResult<T>> {
+    return this.paginatedOperation(path, params, options);
+  }
+
+  /**
+   * Delete some pages from a paginated list, following the `next-page`
```
leplatrem, line 759: I think it would be better phrased with something like `Delete multiple objects, following the pagination if the number of objects exceeds the page limit`

src/collection.ts
```diff
@@ -613,6 +613,55 @@ export default class Collection {
...
+   *
+   * - `{fieldname: "value"}`
+   * - `{min_fieldname: 4000}`
+   * - `{in_fieldname: "1,2,3"}`
+   * - `{not_fieldname: 0}`
+   * - `{exclude_fieldname: "0,1"}`
+   *
+   * @see http://kinto.readthedocs.io/en/stable/api/1.x/filtering.html
+   *
+   * Paginating is done by passing a `limit` option, then calling the `next()`
```
leplatrem, line 636: ...`The default server pagination size can be forced by passing ... `


Also, isn't it automatic? I don't think we should mention the `calling the next() method` no?

paullaffitte:
> > That's very nice that there is some automated tests now. If they were there at the time I started the PR, it would be already merge maybe haha.
> 
> What do you mean? This repo had tests since its first commit :) https://github.com/Kinto/kinto-http.js/tree/87c681c82a

I maybe just didn't remember about it, I was pretty sure that I had to run the tests locally.

I will do the requested changes during this week.

src/collection.ts
paullaffitte: I'm not sure about this, but the `listRecords` function in the same file has a very similar documentation and is using the same `PaginatedParams` object as parameter. If I remove it I think we should also update the documentation of `listRecords`, for consistency sake.

paullaffitte:
Looks like the tests on firefox failed for session issue rather than because of my code (that I didn't change since last successful pipeline anyway, so it would makes sense).

dstaley:
All good now :) I'll review this over the next few days. Thanks again for contributing!

dstaley:
@paullaffitte Thank you so much for contributing, and hanging in there while we reviewed this. I really appreciate your patience and your effort in getting this ready to merge! :tada: :tada:

PR #543, opened 2019-10-28
Add endpoint function to base, batch, bucket and collection

### LINKED ISSUE #127: Expose the endpoint function to client programs

paullaffitte:
Closes #127 
I would like to have feedback on this before writing tests, does it match your expectation?

### DIFF
--- a/src/base.ts
+++ b/src/base.ts
@@ -9,7 +9,7 @@ import {
   cleanUndefinedProperties,
 } from "./utils";
 import HTTP, { HttpResponse } from "./http";
-import endpoint from "./endpoint";
+import endpoint, { get as getEndpoint } from "./endpoint";
 import * as requests from "./requests";
 import { aggregate, AggregateResponse } from "./batch";
 import Bucket from "./bucket";
@@ -872,5 +872,16 @@ export default class KintoClientBase {
         { method: "PUT" }
       )
     );
+  }
+
+  /**
+   * Retrieves a server endpoint by its name.
+   * @param  {String}   name The endpoint name.
+   * @param  {Object} params The endpoint parameters.
+   *
+   * @return {String}
+   */
+  endpoint(name: keyof typeof endpoint = "root", ...args: string[]): String {
+    return getEndpoint(name)(...args);
   }
 }
--- a/src/base.ts
+++ b/src/base.ts
@@ -9,7 +9,7 @@ import {
   cleanUndefinedProperties,
 } from "./utils";
 import HTTP, { HttpResponse } from "./http";
-import endpoint, { get as getEndpoint } from "./endpoint";
+import endpoint from "./endpoint";
 import * as requests from "./requests";
 import { aggregate, AggregateResponse } from "./batch";
 import Bucket from "./bucket";
@@ -875,13 +875,11 @@ export default class KintoClientBase {
   }
 
   /**
-   * Retrieves a server endpoint by its name.
-   * @param  {String}   name The endpoint name.
-   * @param  {Object} params The endpoint parameters.
+   * Retrieves server endpoints.
    *
-   * @return {String}
+   * @return {Object}
    */
-  endpoint(name: keyof typeof endpoint = "root", ...args: string[]): String {
-    return getEndpoint(name)(...args);
+  get endpoint(): Record<string, Function> {
+    return endpoint;
   }
 }
--- a/src/batch.ts
+++ b/src/batch.ts
@@ -1,4 +1,4 @@
-import { get as getEndpoint } from "./endpoint";
+import endpoints from "./endpoint";
 import { KintoRequest } from "./types";
 
 interface ConflictRecord {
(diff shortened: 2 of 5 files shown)


dstaley:
Hi @paullaffitte! Thank you so much for taking a go at this.

One of my main concerns is the ability to retain type-checking on endpoint functions. In the transition to TypeScript, we made each endpoint function specifically type-checkable, and I'd like to make sure that this feature doesn't introduce the ability to pass arguments that pass type-checking but result in invalid parameters.

For example, with this PR as-is the following would be allowed by the type-checker:

```js
const client = new KintoClient('https://kinto.server.com');
const recordUrl = client.endpoint.root("invalid"); // endpoint.root doesn't accept any parameters
```

I think a good first step would be to expose `endpoints` as a property on `KintoClientBase`. That way, the user could do the following:

```js
const client = new KintoClient('https://kinto.server.com');
const recordUrl = client.endpoints.record("itemBucket", "itemCollection", "itemId");
```

This way the type information will be retained, and invalid arities will be caught by the compiler.

I really like how you added in the properties of the instance in the call to `endpoint`, and I definitely think we could add a similar API, but if we add those I think it'd be more appropriate to add explicit functions with correct type signatures rather than a catch-all function.

src/base.ts
```diff
@@ -873,4 +873,13 @@ export default class KintoClientBase {
       )
     );
   }
+
+  /**
+   * Retrieves server endpoints.
+   *
+   * @return {Object}
+   */
+  get endpoint(): Record<string, Function> {
```
dstaley: can you change the return type to `typeof endpoint`? Technically `Function` is a wider type than what we're using.

src/batch.ts
```diff
@@ -93,3 +94,10 @@ export function aggregate(
     return acc;
   }, results);
 }
+
+/**
+ * The server endpoint for batch.
+ *
+ * @return {String}
+ */
+export const endpoint = endpoints.batch;
```
dstaley: Since `KintoClientBase` has the `endpoint` property, I don't think we need this.

src/bucket.ts
```diff
@@ -765,4 +765,23 @@ export default class Bucket {
       aggregate: !!options.aggregate,
     });
   }
+
+  /**
+   * Retrieves bucket server endpoints.
+   *
+   * @return {Object}
+   */
+  get endpoint(): Record<string, Function> {
```
dstaley: Can you create an interface type and use that as the return type? So something like this:

```typescript
interface BucketEndpoint {
  history: () => string;
  group: (group: string) => string;
  // etc
}

File: src/bucket.ts [hunk 4]
```diff
@@ -765,4 +765,23 @@ export default class Bucket {
...
+    return {
+      toString: () => endpoint.bucket(this.name),
+      bucket: () => endpoint.bucket(this.name),
```
dstaley: I think we can remove these two since it would make more sense to have a property directly on `Bucket` representing the URL. I'm not sure what that property should be named, but we can open another issue to discuss that.

src/collection.ts
```diff
@@ -793,4 +793,19 @@ export default class Collection {
       aggregate: !!options.aggregate,
     });
   }
+
+  /**
+   * Retrieves collection server endpoints.
+   *
+   * @return {Object}
+   */
+  get endpoint(): Record<string, Function> {
```
dstaley: Same as with `Bucket`. Please replace this with an interface type.

src/collection.ts
```diff
@@ -793,4 +793,19 @@ export default class Collection {
...
+    return {
+      toString: () => endpoint.collection(this.bucket.name, this.name),
+      collection: () => endpoint.collection(this.bucket.name, this.name),
```
dstaley: Same as with Bucket. Let's remove these and use a property directly on `Collection` to represent the URL.

src/base.ts
```diff
@@ -873,4 +873,13 @@ export default class KintoClientBase {
       )
     );
   }
+
+  /**
+   * Retrieves server endpoints.
+   *
+   * @return {Object}
```
dstaley: Since we have TypeScript return types, you can omit the JSDoc return line. The other functions that have this are leftover from before the TypeScript transition.

src/bucket.ts
```diff
@@ -765,4 +765,23 @@ export default class Bucket {
       aggregate: !!options.aggregate,
     });
   }
+
+  /**
+   * Retrieves bucket server endpoints.
+   *
+   * @return {Object}
```
dstaley: Can be omitted since we have the TS return type.

src/collection.ts
```diff
@@ -793,4 +793,19 @@ export default class Collection {
       aggregate: !!options.aggregate,
     });
   }
+
+  /**
+   * Retrieves collection server endpoints.
+   *
+   * @return {Object}
```
dstaley: Can be omitted since we have the TS return type.

dstaley:
Superseded by #742

PR #562, opened 2019-11-11
Add response types for API methods

### LINKED ISSUE #517: Improve typings for functions that return server data
There are nearly 40 instances where we're returning `unknown` or some variation thereof. Because we know the actual API calls being made, we can improve those typings with the actual API responses.

dstaley:
This PR updates the methods that interact with the API to return the correct data type. This ensures that callers are able to access the expected properties on these methods.

Fixes #517


dstaley:
Superseded by #643

PR #643, opened 2020-01-16
Add response types and emit TypeScript declarations

### LINKED ISSUE #517: Improve typings for functions that return server data

dstaley:
Now that I've confirmed these types work well with how the library is being used in `kinto.js`, I think it's time to enable the declarations! This is an important step to finishing the TypeScript conversion for `kinto.js`.

This PR also removes `"use strict";` since TypeScript emits that by default.

This PR also makes a slight modification to the browser build process. Because we're exporting types for use by the TypeScript compiler in `kinto.js`, Rollup wants to export those types in the browser bundle as well, which causes issues since the browser bundle is a UMD module. This PR adds a new `index.browser.ts` file much like `index.fx.ts` that just re-exports the `KintoClient`. TypeScript 3.8 will introduce type only imports that get removed once passed through the compiler, so this is just a temporary solution until 3.8 is released. I highly doubt we'll cut a non-prerelease prior to the release of 3.8, but I'd rather have the repo in a working state in the interim!

This PR supersedes #562, and fixes #517.


dstaley:
@leplatrem If everything looks good, could you merge and release a new alpha version?

rollup.config.js
```diff
@@ -18,13 +18,16 @@ const geckoBuild = {
       mainFields: ["module", "main", "browser"],
       preferBuiltins: true,
     }),
-    typescript({ include: ["*.ts+(|x)", "**/*.ts+(|x)", "*.js", "**/*.js"] }),
+    typescript({
+      include: ["*.ts+(|x)", "**/*.ts+(|x)", "*.js", "**/*.js"],
+      tsconfigOverride: { compilerOptions: { declaration: false } },
```
itaisteinherz: Maybe create a constant (e.g. `const tsconfigOverride = { compilerOptions: { declaration: false } };`), and then use it here as well as in lines 46-48?

dstaley: The Gecko and Browser builds have different `tsconfigOverride` values, so we'd need to modify the value for the browser build anyway.

dstaley: Switched back to the official rollup plugin, which doesn't need us to override tsconfig, so marking this as resolved.

dstaley:
@leplatrem merged! You can view a diff of the Gecko build [here](https://www.diffchecker.com/E9IZeksN). It's primarily differences between how Babel and TypeScript emit certain features. There are a few actual implementation changes, but those all passed tests so it should be fine!

Let me know how it goes! (Also, if you wouldn't mind, could you do another alpha release so we can get the TypeScript PR for `kinto.js` working on CI? Thank you!)

Issue #470, opened 2019-07-26
Should we shim other browser dependencies?


glasserc:
@ncloudioj raised (in #468) the point that we don't support `Blob` (which is part of the Web platform), which makes it difficult to use this library in Node. That was easy to fix, but there remain other requirements that we don't shim. Should we? Should this be the responsibility of the caller? I have no idea what the "done thing" is so I'm open to suggestions.

dstaley:
The difficulty is making sure that any polyfills don't find their way into browser bundles. One way to make sure of is is to use Rollup to create a node-specific bundle, rewriting global uses of things like `fetch`, `btoa`, and `FormData` to their node-polyfill equivalents. This would be very easy with Rollup's inject plugin, but could potentially complicate using the library in a node project along with a bundler/builder that doesn't load the CommonJS version of the library. However in those circumstances it's likely possible for the user to configure their bundler to load the CommonJS version explicitly, so I think it's a worthwhile tradeoff. Furthermore I think the number of users using kinto-http in a node project along with a bundler is a very small group.

So, my suggestion would be to update the TypeScript/Rollup PR to use Rollup to build the CommonJS bundle, rewriting uses of the globals specified in `setup-jsdom.js` to their node polyfills.

dstaley:
Adding in a new note since I've just burned myself on this issue. `kinto-http` relies on some browser-only things like `FormData` and `Headers` (even using things like `instance of Headers`). Polyfilling in the global scope alleviates any issues, but it might be good to make sure the library itself handles libraries like those in an environment-agnostic way.

That being said, I don't think we should polyfill language features. This just applies to APIs that are browser-only and will remain so.

ncloudioj:
Sounds good to me. We ended up with some other ways to interact with the RS back-end instead of using the Node scripts, so this won't affect our daily tasks with RS any way.

@dstaley Thanks for investigating this! Feel free to close it.

dstaley:
Continuing the discussion started by #678: 

The way I see it, there's three possible solutions:

## 1. What we're doing now

Continue doing what we're doing, and require Node consumers to polyfill things like `fetch` and `FormData` in the global scope. We can recommend libraries in the README, but consumers could use anything that conforms to the interface.

## 2. Provide an easily-importable polyfill

As a slightly nicer alternative, we could provide an easily-importable polyfill (`import 'kinto-http/polyfill'`) that provides the recommended polyfills. This would be easier for consumers, but would still require editing the global scope.

[...]

PR #678, opened 2020-02-19
Remove JSDom from tests

dstaley:
This PR removes JSDom from the tests since it's not necessary (I assume it was at one point, but no longer).

It also adds a note to the README about which polyfills are necessary for running in Node.


dstaley:
@leplatrem what do you think of also adding a polyfill for `Blob`? This would remove our usage of `Buffer` (which is node-only anyway). I say if we're gonna require users to provide _some_ polyfills, we might as well write only browser-safe code and require polyfills for anything outside of that.

This would have the added benefit of removing `Buffer` polyfills when bundled with things like Webpack. For reference, the polyfill injected by Webpack comes in at 19 KB.

leplatrem:
Could there be a way to have polyfills only in Node libs? 

See also https://github.com/Kinto/kinto-http.js/issues/470

dstaley:
closing in favor of https://github.com/Kinto/kinto-http.js/pull/700

PR #690, opened 2020-02-26
Remove chai-as-promised and convert tests to async/await

dstaley:
In preparation for running tests in an actual browser environment, I've refactored all the tests that relied on `chai-as-promised` to async/await. In the process I discovered several tests that weren't waiting on promises correctly, and were throwing unhandled rejections.

Removing `chai-as-promised` was necessary to get the tests passing in [Intern](https://theintern.io/). I'll send a future PR actually adding Intern, but I figured this would be a good PR to merge even without Intern.


dstaley:
@leplatrem would love a quick spot check on this! For the most part it's exactly the same, so just skimming the changes and checking a few spots would be really appreciated!

test/bucket_test.ts
```diff
@@ -42,26 +44,26 @@ describe("Bucket", () => {
...
 
-    it("should resolve with response data", () => {
+    it("should resolve with response data", async () => {
       const response = { data: { foo: "bar" }, permissions: {} };
       sandbox.stub(client, "execute").returns(Promise.resolve(response));
 
-      return getBlogBucket()
-        .getData()
-        .should.become({ foo: "bar" });
+      ((await getBlogBucket().getData()) as { foo: string }).should.deep.equal({
```
leplatrem: Maybe in this case, an intermediary variable would make sense

test/integration_test.ts
```diff
@@ -1274,151 +1185,136 @@ describe("Integration tests", function() {
...
-            .then(({ data }) => {
-              expect(data[0].title).eql("foo");
-              expect(data[0].members).eql(["github:me"]);
-              expect(data[0].blah).eql(43);
-            });
+        it("should update a group", async () => {
+          const res = await bucket.createGroup("foo");
+          await bucket.updateGroup({ ...res.data, title: "mod" });
+          const { data } = await bucket.listGroups();
+          (data[0] as Group & { title: string }).title.should.equal("mod");
```
leplatrem: Oh, i don't know this syntax :grin: ! Could you add a comment please?

test/bucket_test.ts
dstaley: I've updated the tests to use an intermediary variable whenever we were casting an awaited value. However, we still have tests that look like this:

```ts
(await api.fetchServerInfo()).should.deep.equal(fakeServerInfo);
```

Is this clear enough, or should we lift all awaited calls into variables?

leplatrem: Yes, this was like one of the few complex ones. Thanks for the changes!

PR #700, opened 2020-03-03
Remove JSDOM from tests and create polyfilled Node bundle

### LINKED ISSUE #470: Should we shim other browser dependencies?

### LINKED ISSUE #526: Find way to prevent Buffer polyfill from being added by Webpack
When consuming `kinto-http.js` with Webpack, it's seeing [this line](https://github.com/Kinto/kinto-http.js/blob/master/src/utils.ts#L294) and injecting a polyfill for `Buffer` into the bundle, adding 6.5 KB to the gzipped size of the bundle. There's probably a way to circumvent that without sacrificing node.js compatibility.

Looks like `global.Buffer.from(array)` could work.

### LINKED ISSUE #683: Importing NPM module via CommonJS requires `.default`
```
(.venv) PS ~\Desktop\cjs-test> node
Welcome to Node.js v12.15.0.
Type ".help" for more information.
> const KintoClient = require("kinto-http");
undefined
> KintoClient
{ default: [Function: KintoClient] }
```

This is because TypeScript doesn't rewrite default ES6 exports, even when there's only one exported object.

I need to look into how other libraries are solving this, but one possible solution is to use rollup and generate a bundle for Node. I'm not super enthusiastic about this solution.

dstaley:
This PR modifies the Rollup configuration to emit a Node bundle that replaces calls to browser-specific APIs with polyfilled versions. This is accomplished using [`@rollup/plugin-inject`]() which replaces usage in the global scope with an import statement to the polyfill. This PR polyfills `fetch`, `Headers`, `FormData`, `atob`, and `Blob`. The `Blob` polyfill is provided by the `blob.js` file included at the root of the repo.

The third-party polyfills are marked as `external`, meaning they aren't included in the Node bundle, and need to be installed by consumers. They're marked as optional `peerDependencies`, which allows consumers to use any version of the polyfills newer or equal to the latest version at the time of this PR. Instructions are included in the README.

I've also bumped up the minimum Node version to v10. This enables us to ship ES2018 syntax and APIs to Node.

This PR also contains the changes originally included in #609 

Fixes #683, fixes #526, fixes #470


dstaley:
@leplatrem So it looks like Prettier currently doesn't support the new TypeScript 3.8 syntax https://github.com/prettier/prettier/issues/7263

I've added a temporary `.prettierignore` file ignoring `index.ts` until they get things sorted. I've subscribed to the issue, so as soon as it's fixed and a release is made I'll remove it.

test/setup-globals.js
```diff
@@ -0,0 +1,15 @@
...
+
+// jsdom FormData implementation is inconsistent, exposing a better one
+global.FormData = require("form-data");
+
+// atob polyfill for tests
+global.atob = require("atob");
+
+global.Blob = function Blob(dataArray) {
+  return Buffer.from(dataArray[0]);
+};
```
leplatrem: Can't we take advantage of what's in `blob.js` here?

dstaley: Turns out we can! Mocha will load the setup file even if it's TypeScript, so I've updated it to import from `blob.js`.

dstaley:
@leplatrem Can you add me as a package maintainer for `kinto-http.js` on npm so that I can publish an alpha version? That's one of the things blocking https://github.com/Kinto/kinto.js/pull/1172

leplatrem:
Done!

PR #715, opened 2020-03-11
Run tests in browser using Intern

dstaley:
This PR switches from Mocha to Intern in order to facilitate running our entire test suite in an actual browser! :tada:

Due to the fact that the tests are now true ES modules, I had to adjust some stubs to stub/spy on local instances rather than module-level instances. For instance, instead of stubbing `requests.createRequest`, we stub `client.execute`. This has the added benefit of testing the exact HTTP call being made, rather than simply testing the parameters passed to `createRequest`.

Other notable changes include switching from `global` to `globalThis` to support browsers (which don't have a top-level `global`), modifying tests that were passing invalid vales while building requests (such as using `safe: true` without a `last_modified`), and updating integration test options to support different URLs for the Kinto server when run in the browser.

Testing commands retain their original meaning, with three new additions:
- `npm run test:chrome` and `npm run test:firefox` - Runs the tests in Chrome and Firefox respectively
- `npm run build:test-browser` - Bundles the tests for the browser using Rollup and `rollup.test.config.js`

I've also included a new GitHub Actions workflow that tests in both Chrome and Firefox on Windows. I don't suspect the OS to matter that much, but if you disagree I can add Linux and/or macOS to the matrix.

This uses the experimental `0.0.0-experimental-97f199b` version of `kinto-node-test-server`. Once we get this PR merged in and feel confident that the TypeScript version of `kinto-node-test-server` is good to release, I'll release v2.0.0 and update the version in this library.

To make the review simple, I'd love for you to:
- Check out this branch
- Run `npm run test` and verify all tests pass
- Run `npm run test:chrome` and `npm run test:firefox` and verify all tests pass
- Run `npm run test-cover` and check that the coverage report is printed to the console
- Run `npm run test-cover-html` and check that the coverage report opens in your default browser

If you really want to, please feel free to spot check the modified tests. The diff is a lot, but there were hardly any meaningful changes, so I'm pretty confident I didn't mess up anything serious there.


dstaley:
@leplatrem If you get some free time, I'd love for you to take a quick look at this. Since it touches a lot of files (including package.json) it'll generate conflicts pretty quickly, so I'd love to get it merged in.

package.json
```diff
@@ -53,37 +56,36 @@
...
-    "chai": "^4.1.0",
     "coveralls": "^3.0.0",
     "cross-env": "^7.0.0",
     "eslint": "^6.0.1",
     "eslint-config-prettier": "^6.0.0",
     "form-data": "^3.0.0",
     "http-server": "^0.12.0",
-    "kinto-node-test-server": "^1.0.0",
+    "intern": "^4.8.0",
+    "kinto-node-test-server": "0.0.0-experimental-97f199b",
```
leplatrem, line 77: we shoud probably cut a release if this works now. Let me know!

leplatrem:
Feel free to merge, I'll rebase #742 ;)

dstaley:
@leplatrem We can't really get around the second `tsconfig.json` file since we need to pass that into Intern. However, I can merge the second rollup file into the first, and then adjust it so that it only builds the test bundle when `TESTING=true` or something. Would that be better?

leplatrem:
> I can merge the second rollup file into the first, and then adjust it so that it only builds the test bundle when TESTING=true or something. Would that be better?

If we can add the env var in package.json scripts, yes why not :) Otherwise don't worry, as long as we follow the conventions, it's acceptable!

Thanks again!

PR #742, opened 2020-04-06
Expose client.endpoints attribute (fixes #641)

### LINKED ISSUE #641: Support custom resource API endpoints
Kinto internally allow register custom colander based Resource, but i don't see any functionality related to this in kinto-http.js

all kinto-http.js API works only with bucket/collections, but what to do with endpoints  like `/mushrooms/` from this [example](https://docs.kinto-storage.org/en/stable/core/quickstart.html#define-resources)?

If this possible, will be fine document it.

leplatrem:
For Kinto/kinto.js#1751, I was thinking of something like this. Usage would be: 

```js
// Replace record endpoint with custom «resource»
kintoClient.endpoints.record = (bucket: string, coll: string, id?: string) =>
   `/custom/url/records` + (id ? `/${id}` : "");
``` 

```js
// Add new endpoint
kintoClient.endpoints.changeset = (bucket: string, coll: string) =>
   `${client.endpoints.collection(bucket, coll)}/changeset`

const coll = kintoClient.bucket("bid").collection("cid");
const path = coll.endpoints.changeset(coll.bucket, coll.collection);
const response = await coll.client.execute({ path });
```

What do you think?


dstaley:
The first use case (overriding a default endpoint) will work just fine in TypeScript, but the second use case (custom endpoints) might get a bit tricky to represent both accurately and safely as a class property. What if we suggested users create their own `endpoint` object like this:

```ts
const endpoints = {
  ...client.endpoints,
  changeset: (bucket: string, coll: string) =>
    `${client.endpoints.collection(bucket, coll)}/changeset`
};
```

This would allow TypeScript to correctly type all the properties of `endpoints` without introducing some complexity into the classes that have a `_endpoint` property.

leplatrem:
Thanks!

Based on your feedback (and Ethan's in Kinto/kinto-http.js#743 ) I changed my strategy and I don't think we need the use case of adding of a new endpoint.

src/base.ts
```diff
@@ -7,11 +7,11 @@ import {
   cleanUndefinedProperties,
 } from "./utils";
 import HTTP, { HttpResponse } from "./http";
-import endpoint from "./endpoint";
+import Endpoints from "./endpoints";
```
dstaley, line 10: small nitpick: why did you switch from lowercase `endpoints` to uppercase `Endpoints`?

src/bucket.ts
```diff
@@ -29,6 +28,7 @@ export interface BucketOptions {
 export default class Bucket {
   private client: KintoClientBase;
   public name: string;
+  private _endpoints: any;
```
dstaley, line 31: I'd prefer to avoid `any` if at all possible. You _should_ be able to use `KintoClientBase["endpoints"]` as the type.

src/collection.ts
```diff
@@ -34,6 +32,7 @@ export default class Collection {
   public client: KintoClientBase;
   private bucket: Bucket;
   public name: string;
+  private _endpoints: any;
```
dstaley, line 35: Same as above: let's try to avoid using `any`. You _should_ be able to use `KintoClientBase["endpoints"]` here.

test/test_utils.ts
```diff
@@ -1,21 +1,38 @@
 import sinon from "sinon";
 import { expect } from "chai";
 
+class Headers extends Map {
```
dstaley, line 4: Is there a reason you didn't use the existing `Headers` global from [here](https://github.com/Kinto/kinto-http.js/blob/master/test/setup-globals.ts#L6)?

dstaley (changes requested):
Nice work! I really like how you figured out a nicer way to stub out the request method.

test/test_utils.ts
```diff
@@ -1,21 +1,24 @@
...
 export function fakeServerResponse(
   status: number,
   json: any,
   headers: { [key: string]: string | number } = {}
 ) {
+  const respHeaders = fakeHeaders(headers);
+  if (!respHeaders.has("Content-Length")) {
+    respHeaders.set("Content-Length", JSON.stringify(json).length);
+  }
   return Promise.resolve({
```
dstaley, line 19: Since TypeScript already has a global declaration for `Headers`, we don't need to pull it off of `global`. Also, by casting `global` to `any`, the resulting `Headers` is _also_ any, meaning the calls to `has` and `set` also lose typechecking.

I tried to propose this as a suggested change, but the diff was difficult to follow since the line numbers are affected by Prettier's formatting. I also didn't want to risk giving you something that ends up failing the formatting checks!

I've marked the changed lines with an asterisk:

```ts
export function fakeHeaders(headers: { [key: string]: string | number } = {}) {
  const h = new Headers(); // * don't cast global as any
  Object.entries(headers).forEach(([k, v]) =>
    h.set(k, typeof v === "number" ? v.toString() : v) // * convert numbers to strings
  );
  return h;
}

export function fakeServerResponse(
  status: number,
  json: any,
  headers: { [key: string]: string | number } = {}
) {
  const respHeaders = fakeHeaders(headers);
  if (!respHeaders.has("Content-Length")) {
    respHeaders.set("Content-Length", JSON.stringify(json).length.toString()); // * convert number to string
  }
  return Promise.resolve({
    status: status,
    headers: respHeaders,
    text() {
      return Promise.resolve(JSON.stringify(json));
    },
  });
}
```

dstaley (changes requested):

glasserc (approved):
LGTM but Dylan's review should supersede mine :)

leplatrem:
Thanks for your help @dstaley! I think we're good here, I'll merge :)

PR #743, opened 2020-04-06
Add changeset endpoint

leplatrem:
In [kinto-changes](https://github.com/Kinto/kinto-changes/#endpoints) we added a new `.../changeset` endpoint. 

@glasserc do you think I could add it here? Or I should improve #742 to support that use-case?

Thanks for your feedback :)

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -800,4 +800,21 @@ export default class Collection {
       aggregate: !!options.aggregate,
     });
   }
+
+  @capable(["changes"])
+  async getChangeset<T>(
+    options: {
+      headers?: Record<string, string>;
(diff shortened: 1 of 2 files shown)


glasserc:
I definitely think #742 is a nicer, more extensible approach to this problem. The server is built around the concept of plugins, so I guess the client should be too, right?

leplatrem:
With what I propose in #742, it could become:

```js
const path = client.endpoints.changeset(bid, cid);
await client.execute(path, { query: { _expected: 42 } });
```

I think it's better :) `kinto-changes` could be renamed `kinto-remotesettings` and I don't think this should go in the basic kinto http client

PR #760, opened 2020-04-16
Expose execute in bucket and collection classes

leplatrem:
Instead of having:

```js
const client = new KintoClient(..);
const clientColl = client.bucket(bid).collection(cid);

await clientColl.list();
await client.execute(...);

```

We could just have one reference to the collection instance and still execute raw requests.

### DIFF
--- a/src/base.ts
+++ b/src/base.ts
@@ -518,7 +518,6 @@ export default class KintoClientBase {
   /**
    * Executes an atomic HTTP request.
    *
-   * @private
    * @param  {Object}  request             The request object.
    * @param  {String}  request.path        The path to fetch, relative
    *     to the Kinto server root.
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -68,6 +68,19 @@ export default class Bucket {
     this._safe = !!options.safe;
   }
 
+  async execute<T>(
+    request: KintoRequest,
+    options: {
+      raw?: boolean;
+      stringify?: boolean;
+      retry?: number;
+      query?: { [key: string]: string };
+      fields?: string[];
+    } = {}
+  ): Promise<T | HttpResponse<T>> {
+    return this.client.execute(request, options);
+  }
+
   get headers(): Record<string, string> {
     return this._headers;
   }
--- a/src/bucket.ts
+++ b/src/bucket.ts
@@ -68,17 +68,8 @@ export default class Bucket {
     this._safe = !!options.safe;
   }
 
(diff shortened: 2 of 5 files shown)


dstaley:
I haven't tried it, but I think you can simplify these down to getters, which avoids having to redeclare the parameter types:

```ts
get execute() {
  return this.client.execute;
}
```

[Here's an example](https://www.typescriptlang.org/v2/en/play?#code/MYGwhgzhAEBCkFNoG8BQ1oAcCuAjEAlsNAgB4LDYAuCAPACoB8AFAPaZUGsB2EAXCmgAnMAHcBuVqxAIw3ADTCEVIQE8B3bAFtcCIdAC+ASgEAFIay0EIdetAA+0egG0AuoxToM0AgDNobBxcvAB0IqJGnt7eQsrYQtzQ5pbWCGEIENIAbgjMyAbQkE5GANxeGAao5UpU8YnJVjbpmSA5zM75hTD0rqVelZWooJAwAMKECNxUUVhCBFlgNNC4iALwNmVewDwQKtjAVKxCzJFo0VQAFtYhKzbQALzQ3AiicIgnZRVVGADmyiTkSg0E4zDCxWoJaCXa63NJkCjUBCfQyoQbbXjTYiPZ6vcYESZUD6oZiQVTcYgg+4eZDQLwAejpwh8ME0Oj0Die2l0QjcWx2030jzEYAImJC8KBdFZ3JYNPCAl8YBANkU4LUAgADIY+vTGSBmdBdnNuD8OUaCCbeRh0btoPqhaIRWKJYjaOaTbLhGIBHsEKrlOroFrjGVjB8gA). You can see that the correct type is inferred by hovering over the return type.

leplatrem:
It doesn't seem to work with eslint, and I can't get my way out of it :/

* `Missing return type on function.eslint@typescript-eslint/explicit-function-return-type`
* Adding ` /* eslint-disable eslint@typescript-eslint/explicit-function-return-type */`
* `Definition for rule 'eslint@typescript-eslint/explicit-function-return-type' was not found.`
* Adding `get execute<T>(): Promise<T | HttpResponse<T>> {`
* `An accessor cannot have type parameters.ts(1094)`

... :man_shrugging:

dstaley:
Ah, sorry, this is what I get for suggesting things before trying them myself!

Looks like this is what we want:

```ts
get execute(): KintoClientBase["execute"] {
  return this.client.execute;
}
```

PR #956, opened 2021-02-12
Fix implementation of collection snapshot

leplatrem:
The previous implementation was not handling properly deletions/recreations. This one is much simpler and works as expected.

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -814,7 +814,6 @@ export default class Collection {
         recordsById.set(record.id, record);
       }
     }
-    console.debug(`Found ${recordsById.size} records in snapshot (at=${at}).`);
     return {
       last_modified: String(at),
       data: Array.from(recordsById.values()).sort(
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -787,7 +787,6 @@ export default class Collection {
         "max_target.data.last_modified": String(at), // eq. to <=
       },
     });
-    console.log(`${changes.length} history entries found.`);
     return changes;
   }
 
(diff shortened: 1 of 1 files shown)


src/collection.ts
```diff
@@ -780,13 +780,14 @@ export default class Collection {
...
       pages: Infinity, // all pages up to target timestamp are required
-      sort: "-target.data.last_modified",
+      sort: "target.data.last_modified",
       filters: {
         resource_name: "record",
         collection_id: this.name,
         "max_target.data.last_modified": String(at), // eq. to <=
       },
     });
+    console.log(`${changes.length} history entries found.`);
```
dstaley, line 790: We should probably figure out a system to allow for optional debugging statements, but for now let's remove calls to `console.log` and `console.debug`.

dstaley (changes requested):
Looks great! Definitely easier to read.

src/collection.ts
```diff
@@ -801,30 +801,30 @@ export default class Collection {
...
-        snapshot = snapshot.filter((r) => r.id !== record.id);
-      } else if (!seenIds.has(record.id)) {
-        seenIds.add(record.id);
-        snapshot.push(record);
+        recordsById.delete(record.id);
+      } else {
+        recordsById.set(record.id, record);
       }
     }
+    console.debug(`Found ${recordsById.size} records in snapshot (at=${at}).`);
```
dstaley, line 817: Should we keep this?

dstaley (approved):
I'll leave it up to you if you want to remove the call to `console.debug`.

src/collection.ts
leplatrem: No, I agree, this shouldn't be the first one

PR #964, opened 2021-02-17
Fix #963: Fix snapshot when records were deleted via plural endpoint

### LINKED ISSUE #963: Snapshot of collection is incorrect if records deleted via the plural endpoint
Because of https://github.com/Kinto/kinto/issues/1649, the records that are deleted via the plural endpoint are not correctly handled in the snapshot.

**Steps to reproduce**

```js
await coll.createRecord({ n: 1 });
await coll.createRecord({ n: 2 });
await coll.createRecord({ n: 2 });
await coll.deleteRecords({
  filters: {
    eq_n: 2,
  }
});
const { data: rec3 } = await coll.createRecord({ n: 3 });
const { data } = await coll.listRecords({
  at: rec3.last_modified,
});
```

**Expected**

```js
[{ n: 3 }, { n: 1 }]
```

**Actual**

`[{ n: 3 }, { n: 2 }, { n: 2 }, { n: 1 }]`

leplatrem:
Unfortunately the fix in #956 was not enough to cover the misbehavior of the Admin UI diff view. I traced it back and it was related to the fact that plural deletes do not generate history entries.

This approach seems to work as expected, and does not introduce too much performance loss

Fixes #963

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -817,6 +817,7 @@ export default class Collection {
     // Current records ids in the collection.
     const { data: current } = await this.listRecords({
       pages: Infinity,
+      fields: ["id"], // we don't need attributes.
     });
     const currentIds = new Set(current.map((record) => record.id));
 
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -11,7 +11,6 @@ import {
   KintoIdObject,
   KintoObject,
   Attachment,
-  HistoryEntry,
   OperationResponse,
   MappableObject,
 } from "./types";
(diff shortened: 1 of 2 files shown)


src/collection.ts
```diff
@@ -768,7 +767,14 @@ export default class Collection {
   /**
    * @private
    */
-  async listChangesUntil<T>(at: number): Promise<HistoryEntry<T>[]> {
+  @capable(["history"])
```
Natim, line 770: What is this syntax?

leplatrem: https://docs.google.com/document/d/1ikxIP5-RVYq6d_f8lAvf3pKC00W78ueyp-xIZ6q67uU/edit

PR #1191, opened 2021-11-23
Address security vulns

matt-boris:
Found 3 new vulns here https://github.com/Kinto/kinto-http.js/security/dependabot


leplatrem (approved):
Oh, so we list them as top level dependencies?

matt-boris:
> Oh, so we list them as top level dependencies?

@leplatrem yeah, so that the newer version is used instead of the version brought in by the dependency. Check out [this](https://github.com/Kinto/kinto-http.js/security/dependabot/package-lock.json/bl/open) and their example of remediating.

PR #1267, opened 2022-02-16
Restore Windows integration tests

leplatrem:
Follow up of https://github.com/Kinto/kinto-http.js/pull/1266

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -1,0 +1,30 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+name: CI
+jobs:
+  chore:
+    name: Windows Integration Tests
+    runs-on: windows-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


dstaley:
Looks like Kinto v4.6.1 was published on Feb 3. Might be worth investigating if that release has a Windows specific breakage.

dstaley:
I ran the test suite for kinto v14.6.1 on Windows, and the only failing test was `tests/core/test_storage.py::PostgreSQLStorageTest::test_the_timestamp_are_based_on_real_time_milliseconds`, which I can almost guarantee is just due to how timers work on Windows, so it's likely not our culprit here.

dstaley:
Turns out we were upgraded to a new Windows image in Github Actions between those two workflow runs. You can tell since `pip` was updated to `22.0.2`, so I wonder if this isn't actually a `kinto` issue and is more something going wrong with pip.

Additionally, I'm able to run Windows integration tests without issue using `pip 19.2.3`, so this seems likely.

